PR #83, opened 2023-04-30
Add line height property for drop caps

Nazaretto:
Some of old slavic fonts (like Vilnius) have inconsistencies line height and font size, which in case of drop caps which causes some positioning problems. The easiest fix for that is to allow adjusting line height manually.

Screen Shot 2023-04-30 at 15 33 56


danielgarthur:
Congratulations on being our first contributor to make a pull request! In general the approach looks correct, I'll leave a few comments.

danielgarthur:
My general thoughts are 

- The dropcap element would have a `lineHeight` property that is a nullable number that represents the unitless number described [here](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height).
- If `lineHeight` is `null`, then the CSS should use `normal`. Or just omit the property in this case because the default value is `normal`.
- Use `InputUnit` with type `unitless` as the input, with an appropriate min, max, step size, and precision. This component does not current support blank values, but could be extended to have `nullable` prop that allows blank values.

src/services/integration/ByzHtmlExporter.ts
```diff
@@ -150,6 +150,7 @@ export class ByzHtmlExporter {
         --byz-drop-cap-font-size: ${Unit.toPt(
           pageSetup.dropCapDefaultFontSize,
         )}pt;
+        --byz-drop-cap-line-height: ${pageSetup.dropCapDefaultLineHeight};
```
danielgarthur, line 153:
```suggestion
```

There is no `--byz-drop-cap-line-height` in byzhtml. Instead, dropcaps always have `lineHeight: 0` and a vertical offset is applied. In Neanes, drop caps are positioned absolutely, and so line height does not affect the HTML flow. But byzhtml is built to allow drop caps to be used with any kind of positioning. The method `getDropCapAdjustment` should calculate the vertical offset correctly. This will need be tested, but since you already added the new line height property to the calculation, it will probably work.

src/services/SaveService.ts
```diff
@@ -553,6 +554,7 @@ export class SaveService {
       p.dropCapDefaultFontStyle ?? pageSetup.dropCapDefaultFontStyle;
     pageSetup.dropCapDefaultStrokeWidth =
       p.dropCapDefaultStrokeWidth ?? pageSetup.dropCapDefaultStrokeWidth;
+    pageSetup.dropCapDefaultLineHeight = p.dropCapDefaultLineHeight;
```
danielgarthur, line 557:
```suggestion
    pageSetup.dropCapDefaultLineHeight = p.dropCapDefaultLineHeight ?? pageSetup.dropCapDefaultLineHeight;
```

It's best to fall back here to the default page setup line height to keep the existing code pattern.

src/components/ToolbarDropCap.vue
```diff
@@ -24,6 +24,12 @@
         @input="$emit('update:fontSize', $event)"
       />
       <span class="space"></span>
+      <input
```
danielgarthur, line 27: Use `InputUnit` with type unitless and all the appropriate props.

`InputUnit will need to be expanded to handle "blank" e.g. nullable values. If you need help with this part, let me know.

src/components/PageSetupDialog.vue
```diff
@@ -235,7 +235,13 @@
               v-model="form.dropCapDefaultFontSize"
             />
           </div>
-
+          <div class="form-group">
+            <label class="drop-caps-label">Line height</label>
```
danielgarthur, line 239:
```suggestion
            <label class="drop-caps-label">Line Height</label>
```

For UI consistency, Height should be capitalized.

src/components/PageSetupDialog.vue
```diff
@@ -235,7 +235,13 @@
...
+            <input
```
danielgarthur, line 240: See comments about `InputUnit` in the toolbar file.

danielgarthur (changes requested):

Nazaretto:
I've tried to implement `nullable` state for `InputUnit` with least changes possible, although I'm not familiar with all nuances of writing components with Vue, so if you have any suggestion, let me know.

src/components/InputUnit.vue
```diff
@@ -51,12 +52,27 @@ export default class InputUnit extends Vue {
   get displayValue() {
     let convertedValue = this.toDisplay(this.value);
 
+    if (convertedValue === null || convertedValue === undefined)
```
danielgarthur, line 55:
```suggestion
    if (convertedValue == null)
```

should work and would be consistent with the existing code style. `== null` checks for both `null` or `undefined` at the same time.

src/components/InputUnit.vue
```diff
@@ -51,12 +52,27 @@ export default class InputUnit extends Vue {
...
+      return this.nullable ? '' : '0';
```
danielgarthur, line 56: The latest `master` branch now has a `default`  value that will need to be used here instead of 0. Try merging in latest `master`.

src/components/InputUnit.vue
```diff
@@ -51,12 +52,27 @@ export default class InputUnit extends Vue {
...
+  emitValue(v: number | null) {
+    if (this.value !== v) {
+      this.$emit('input', v);
+    } else {
+      this.htmlElement.value = this.displayValue;
+    }
+  }
+
   onChange(input: string) {
+    if (input === '' && this.nullable) {
```
danielgarthur, line 72:
```suggestion
    if (input.trim() === '' && this.nullable) {
```

Just to be safe. Someone might try to clear the input but accidentally leave a space.

src/models/save/v1/Element.ts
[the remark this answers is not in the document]
danielgarthur: This still needs to be done.

danielgarthur (changes requested):

danielgarthur (approved):

PR #91, opened 2023-05-16
`git status` is not clean after build

### LINKED ISSUE #99: `git status` is not clean after following instructions on a new computer
### Steps to reproduce

- Follow the instructions to build the project in the `README`.
- Run `git status`.

### Expected results

`git status` is clean.

### Actual results

`git status` shows a dirty `package-lock.json`.

### Notes

In https://github.com/danielgarthur/neanes/pull/92#issuecomment-1550253036 @danielgarthur wrote:

> My expectation would be that developers use `npm install`. If this results in a change to `package-lock.json`, then this would committed as part of the PR. But there should never be a PR that is purely to update package-lock.json.

This would make sense if and only if there was a rule that any change to `package.json` must also be accompanied by running `npm install` and checking in the updated `package-lock.json`. But such a rule is not in place, as evidenced by commit 58fcde4. The fact that such a rule is not in place means that someone making an unrelated change (or simply running the build on a new computer) will end up with unrelated changes to `package-lock.json` and a broken `git status`, which in turn will result in `git commit -a` being broken and likely negative feedback in code reviews. _Valid_ solutions to this problem include:

- #91 Continue to use `npm install` as the supported mechanism in the `README`, but _also_ require that any changes to `package.json` be accompanied by `npm install` and checking in an updated `package-lock.json`. This ensures that `git status` is always clean.
- #92 Maintain the existing laxity around updating `package-lock.json` but change the `README` to recommend `npm ci` instead.

> I would consider this standard practice

It is not the standard practice in another project I maintain which has significantly more developers than this one.

basil:
### Environment

Ubuntu 22.04.2 LTS x86_64
Node v16.20.0
NPM 8.19.4

### Steps to reproduce

```bash
$ git checkout 26059a0  # HEAD
$ git clean -fxd
$ npm install
$ git status
```

### Expected results

`git status` is clean.

### Actual results

`git status` shows:

```
$ git status
[…]
modified:   package-lock.json
[…]
```

### Evaluation

`package-lock.json` was not checked in after a recent modification to `package.json`.

### Solution

Run `git add package-lock.json`.

### Testing done

```bash
$ git clean -fxd
$ git status  # clean
$ npm install
$ git status  # clean
$ npm test
$ npm run electron:build
$ ./dist_electron/Neanes-0.3.0-beta.19.AppImage  # success
```

Fixes #99.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,7 @@ jobs:
           path: ./node_modules
           key: ${{ runner.os }}-modules-${{ hashFiles('package-lock.json') }}
 
+      - run: corepack enable npm
       - name: Install dependencies
         if: steps.cache.outputs.cache-hit != 'true'
         run: npm ci
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -25,6 +25,7 @@ jobs:
           path: ./node_modules
           key: ${{ runner.os }}-modules-${{ hashFiles('package-lock.json') }}
 
+      - run: corepack enable npm
       - name: Install dependencies
         if: steps.cache.outputs.cache-hit != 'true'
         run: npm ci
--- a/.github/workflows/publish-web.yml
+++ b/.github/workflows/publish-web.yml
@@ -24,6 +24,7 @@ jobs:
           path: ./node_modules
           key: ${{ runner.os }}-modules-${{ hashFiles('package-lock.json') }}
 
+      - run: corepack enable npm
       - name: Install dependencies
         if: steps.cache.outputs.cache-hit != 'true'
         run: npm ci
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -27,6 +27,7 @@ jobs:
           path: ./node_modules
           key: ${{ runner.os }}-modules-${{ hashFiles('package-lock.json') }}
 
+      - run: corepack enable npm
       - name: Install dependencies
         if: steps.cache.outputs.cache-hit != 'true'
         run: npm ci
--- a/README.md
+++ b/README.md
@@ -43,13 +43,14 @@ To learn how to use the software, read the [guide](https://danielgarthur.github.
 
 ### Prerequisites
 
-- Node.js 16
+- Node.js 16.9.0 or later
 
 ### Project Setup
 
 First install the project dependencies:
 
 ```
+corepack enable npm
 npm install
 ```
 
--- a/README.md
+++ b/README.md
@@ -73,6 +73,8 @@ npm run electron:serve
 
 This will launch the application in development mode. As changes are made to the source code, the app will reload automatically.
 
+Any change that updates `package.json` _must_ include the corresponding update to `package-lock.json` after running `npm install`.
+
 ### Code Formatting
 
 To format the code before committing, type:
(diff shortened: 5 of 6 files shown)


danielgarthur:
What is the long-term goal here? `package-lock.json` will change on an almost daily basis whenever `npm install` is run because deep dependencies are updated often.

basil:
The long-term goal is [reproducible builds](https://reproducible-builds.org/), which is needed in order to collaborate with others. I suggest you read [the NPM documentation](https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json):

> `package-lock.json` is […] intended to be committed into source repositories, and serves various purposes […]

danielgarthur:
I'm not following. I'm familiar with lock files, but I'm not sure what you are wanting to change. If you want to use the lock file without modification, use `npm ci`.

basil:
Reopening. Either this PR or or #92 should be accepted to fix #99. The status quo, in which #99 exists, is not a viable option.

basil:
The reopened version of this PR is framed in terms of a change to the development rules and a retroactive implementation of that change. See the commit messages for details.

basil:
If this approach is accepted (as opposed to #92), I will file a follow up PR to enforce the new rule by running `npm install` in CI and failing any PRs that update `package.json` without also updating `package-lock.json`.

danielgarthur:
I was able to reproduce the differences in files with docker using the your version and mine (16.20.0 and 16.3.1).

I poked around some other projects to see what they are doing (VS Code, Beekeeper Studio, Azure Data Studio were three that came to mind). Their developer setup does not use `npm ci` or the `yarn` equivalent.  So far, the main difference I see are that

1. They use `yarn`. But I'm pretty sure the `yarn.lock` would get updated when following their dev setup readmes. I would have to test this. But if yarn is better in this regard, than we can switch to yarn.
2. They use `.nvmrc` to lock the node version. 
3. They use bots to update the yarn/package lock dependencies. So maybe they are keeping them up-to-date that way?

I didn't have to time to dive deeper, so I'm not sure if there are other steps that should be taken.

basil:
I looked into this further and determined the Node version does not matter, but the NPM version does. So you are using NPM version 8.1.2 and I am using NPM 8.19.4, hence the difference.

Although we use Yarn rather than NPM in Jenkins, we pin the version of Yarn manager in `package.json` with https://nodejs.org/dist/latest-v16.x/docs/api/all.html#all_packages_packagemanager and the same can be done here. See also https://nodejs.org/dist/latest-v16.x/docs/api/corepack.html for more documentation about corepack.

After pinning the package manager in `package.json`, the corepack workflow works for me:

```
$ git status
[…]
nothing to commit, working tree clean
$ git clean -fxd
$ node -v
v16.20.0
$ npm -v
8.19.4
$ corepack enable npm
$ npm -v
8.1.2
$ npm install
[…]
$ git status
[…]
nothing to commit, working tree clean
$ 
```

Corepack (a package manager for package managers) was added in v16.9.0 / v14.19.0, so this PR is now enabling corepack for this repository in the documentation and CI builds. Accordingly, I have updated the documentation to mention the requirement of 16.9.0 or later and updated the CI builds to run with corepack. As such, the latest version of this PR no longer touches `package-lock.json`, but it does fix #99.

In the long term I will file additional PRs to keep this and other dependencies up-to-date automatically using bots. When I am done with this repository, everything will be kept up-to-date through automation.

From my perspective, this PR is now ready to land as a first step in that direction.

danielgarthur (approved):

PR #100, opened 2023-05-17
Upgrade NPM from 8.1.2 to 8.19.4

basil:
This PR upgrades NPM to a more recent version. Since the newer version has various bug fixes, the generated `package-lock.json` is now different. Since the version is now specified in `package.json`, other developers who are using Corepack (as recommended in the `README`) as well as CI builds should get the new version after pulling these changes. Tested by building the application and opening my Axion Estin score successfully.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6621,127 +6621,6 @@
       "dev": true,
       "optional": true
     },
-    "node_modules/boxen": {
-      "version": "5.1.2",
-      "resolved": "https://registry.npmjs.org/boxen/-/boxen-5.1.2.tgz",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -43280,7 +42239,7 @@
       "requires": {
         "@vue/cli-shared-utils": "^4.5.6",
         "chokidar": "^3.0.2",
-        "electron-builder": "^22.2.0",
+        "electron-builder": "^23.6.0",
```
basil, line 42242: The only line added in this automatic update. This seems like a desirable bug fix, because we explicitly specify this version in `package.json`.

PR #109, opened 2023-05-22
Remove unused `core-js` dependency

basil:
I could not find any usages of this old dependency, so keeping it in the dependency tree seems like a liability. Therefore I am removing it in this PR. If it does need to be retained for some reason, I can close this PR and open a new PR to upgrade it to the latest version instead. Tested with

```
$ npm install
$ npm test
$ npm run electron:build
```

and opening my Axion Estin score successfully in the resulting AppImage build.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,6 @@
       "license": "GPL-3.0",
       "dependencies": {
         "@types/mime-types": "^2.1.1",
-        "core-js": "^3.4.3",
         "html-to-image": "^1.11.11",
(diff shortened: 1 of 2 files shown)


danielgarthur:
From what I can tell, this is a dependency of Babel and is also marked as an optional peer dependency, so that users can choose a different version of `core-js` than the one specified in Babel's dependency tree. So I think it's safe to remove this from our `package.json`, and it will be installed by Babel.

danielgarthur (approved):

PR #111, opened 2023-05-23
Upgrade ESLint from 2019 vintage to 2021 vintage

basil:
Upgrade ESLint from a 2019 vintage to a 2021 vintage release, which also entailed suppressing a seemingly spurious warning introduced in the new version. Tested with

```
$ npm install
$ npm test
$ npm run electron:build
```

and opening my Axion Estin score successfully in the resulting AppImage build as well as intentionally introducing a misspelled variable to make sure it was caught properly by the new version of ESLint.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,14 +43,14 @@
         "@vue/cli-plugin-router": "^4.5.19",
         "@vue/cli-plugin-typescript": "^4.5.19",
         "@vue/cli-service": "^4.5.19",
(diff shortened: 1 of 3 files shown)


danielgarthur:
I recall having trouble with that line a while ago and I added `as EventListener` after finding this [Github issue](https://github.com/microsoft/TypeScript/issues/28357#issuecomment-436484705). There may be other ways to make eslint stop complaining, but they're probably not worth it since it has no effect on the compiled code. I'm good with the disabling the eslint error for that line.

PR #161, opened 2023-07-11
Add `vite-plugin-vue-devtools`

basil:
This doesn't seem like it could hurt, and it might be helpful during debugging I suppose. References:

- https://github.com/webfansplz/vite-plugin-vue-devtools
- https://gist.github.com/webfansplz/bc90a773a0dd474a34e043ab2d2a37a4

To test this I verified that the developer tools worked when running `npm run dev` but did not show up when running a production build.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -51,6 +51,7 @@
         "vite-plugin-electron-renderer": "^0.14.5",
         "vite-plugin-eslint": "^1.8.1",
         "vite-plugin-pwa": "^0.16.4",
+        "vite-plugin-vue-devtools": "^0.5.1",
         "vitepress": "^1.0.0-beta.5",
         "vitest": "^0.33.0",
         "vue-tsc": "^1.8.4"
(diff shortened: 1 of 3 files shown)


danielgarthur:
Is this a replacement for electron-devtools-installer?

basil:
It looks like it. I don't feel strongly about including it one way or another, though.

danielgarthur:
Sorry, what I should have said was that the electron-devtools-installer is already being used. It's useful. If this does the same thing but keeps the code out of our main process code, that sounds like a good thing.

basil:
Oh wow, I did not realize `electron-devtools-installer` was already in place. I was misreading that as `electron-installer` rather than `electron-devtools-installer`. Would not have bothered to open this PR otherwise. I suppose we could possibly rip out `electron-devtools-installer` in favor of this newer extension, but I can't judge if that would be better or worse than the status quo overall. Keeping the code out of the main process code could be one advantage, but I'm not sure if there are other disadvantages. I'm inclined to close this PR.

PR #252, opened 2023-08-29
Fix martyria calculation for most multi-note neume groups

basil:
Fix martyria calculation for most compound neume groups. For neume groups that are compound in the horizontal direction (e.g., a running elaphron or two neumes above a supporting oligon) this is trivial: apply the fthora to the particular portion of the neume group that the fthora is on.

For neume groups that are compound in the vertical direction (e.g., an hyporoe or a non-supporting oligon) things are more ambiguous, so we quickly read through https://analogion.com/psaltologion/neoklis/Tameion_Xourmouziou_A.pdf (a book of very high editorial quality) and made observations about three cases:

1. Hyporoe: Some interesting examples can be found in Manuel's "Makarios Anhr", Bereketis' "Theotoke Parthene", and Chourmouzios' Enharmonic Grave Mode Doxology. Consistent throughout is that when the fthora is on top of the hyporoe it applies to the first part of the hyporoe, while when the fthora is under the hyporoe it applies to the second part of the hyporoe. This makes sense as a general rule, and implementing it this way satisfies the dozen or so examples in the book.
2. Kentemata + oligon: The fthora seems to always be on top and applies to the oligon rather than the kentemata. Rather than imagine what might happen if the fthora was underneath the oligon, we decided to unconditionally apply it to the oligon. This satisfies all of the half dozen or so examples in the book.
3. Oligon + kentemata: When the fthora is on the bottom, it always applies to the oligon. In the book there are several examples where the fthora is written on top and applies to the oligon, but there are also several examples where the fthora is on top and applies to the kentemata. For now we are assuming that if it is on top it applies to the kentemata pending #253.

We feel about 95% confident about our handling of the hyporoe but only about 80% confident about our handling of the kentemata plus oligon and the oligon plus kentemata. But even in its current form, we feel this change is an improvement over the status quo in that it correctly handles the dozen or two examples we found in the abovementioned book. If additional information comes to light in another book of similar editorial quality than we can make further improvements accordingly.

### DIFF
--- a/src/models/NeumeValues.ts
+++ b/src/models/NeumeValues.ts
@@ -135,11 +135,7 @@ export function getSpreadIndex(
     case QuantitativeNeume.OligonPlusKentemataPlusHypsiliRight:
     case QuantitativeNeume.OligonPlusKentemata:
       if (neumeSelection === NeumeSelection.Primary) {
-        if (
-          fthora.endsWith('_Top') ||
-          fthora.endsWith('_TopSecondary') ||
-          fthora.endsWith('_TopTertiary')
-        ) {
+        if (fthora.endsWith('_Top')) {
           return 1;
         } else if (fthora.endsWith('_Bottom')) {
           return 0;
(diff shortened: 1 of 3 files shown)


basil:
I made some more improvements to this PR after publishing it. It still doesn't handle the first case from #253 due to lack of a neume select feature as described in that ticket, but otherwise it handles every other case I could find in that 594 page book, which is quite a few cases that are not currently handled correctly today.

PR #303, opened 2023-09-28
Support ison playback on rests

basil:
The software allows you to place an ison on a rest. I have no idea why anybody would want to do this, but if it is to be allowed then playback should be consistent with the user experience and play it back correctly. I tested this interactively.

### DIFF
--- a/src/services/audio/AudioService.ts
+++ b/src/services/audio/AudioService.ts
@@ -140,6 +140,27 @@ export class AudioService {
         this.toneEvents.push(toneEvent);
       } else if (event.type === 'rest') {
         const toneEvent = new ToneEvent((time) => {
+          const isonUnison = event.isonFrequency === -1;
+
+          if (isonUnison) {
+            if (event.frequency != null) {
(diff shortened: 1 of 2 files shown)


src/services/audio/AudioService.ts
```diff
@@ -140,6 +140,27 @@ export class AudioService {
...
+          } else if (event.isonFrequency !== currentIsonFrequency) {
+            currentIsonFrequency = event.isonFrequency;
+
+            if (event.isonFrequency != null) {
+              isonSynth.triggerAttack(event.isonFrequency!, time);
+            } else {
+              console.warn('AudioService: missing ison frequency', event);
+            }
+          }
+
```
basil, line 163: Copypasta from a few lines above. I tried and failed to abstract this into a common function.

danielgarthur:
Allowing ison on a rest was an oversight on my part. If you've never seen such a thing before, then I would rather remove the ability to put an ison on the rest. I can't recall seeing such a thing before.

basil:
Closing in favor of #308.

Issue #318, opened 2023-10-02
Sharp above and flat below hyporoe


basil:
Currently a flat may be placed on top of an hyporroe and is played back on the second note of the spread. This seems wrong, as in Pringos's Axion Estin in Acem-Kürdî a flat placed above an hyporroe is clearly intended to apply to the first note of the spread. Moreover, when an enharmonic ajem is placed on an hyporroe, the software (correctly) allows it to be placed either above or below the hyporroe and its effect correspondingly applies to either the first note or second note of the spread. The same should be true of flats and sharps I think.

![hyporroe](https://github.com/danielgarthur/neanes/assets/29850/20af7859-3bb1-4852-bc09-35e1649068f2)

basil:
The software currently allows the following:

- Flat above hyporoe (caveat: it doesn't look pretty)
- Sharp below hyporoe - sharp applies to second note of the spread

This ticket covers the following scope:

- Make flat above hyporoe look pretty (see samples below) - flat applies to first note of the spread
- Add support for sharp above hyporoe (see sample below) - sharp applies to first note of the spread
- Add support for flat below hyporoe (I could not find a sample, but I'm sure this exists somewhere) - flat applies to second note of the spread

**Samples from Arvanitis:**

Flat above hyporoe:

![flat-hyporoe-above](https://github.com/danielgarthur/neanes/assets/29850/d08c4204-f0df-4289-ad46-9d6b4d96d772)
![flat-hyporoe-above2](https://github.com/danielgarthur/neanes/assets/29850/4e5dfadf-e41c-40b6-9248-be8fd3976b60)

Sharp above hyporoe:

![sharp-hyporoe-above](https://github.com/danielgarthur/neanes/assets/29850/0248f25b-c3bd-4d69-9352-ab5c3f76345a)

basil:
https://github.com/danielgarthur/sbmufl/pull/14 makes the flat above the hyporoe look pretty; the rest of this ticket remains.

PR #322, opened 2023-10-04
Improve playback service

basil:
This PR is a significant revision of the playback subsystem. While I had originally intended to refactor the portions of the layout service and the playback service that interpret neumes into a separate analysis service, this didn't turn out as well as I had hoped in practice due to the largely different concerns of layout versus playback. So instead I kept the core idea, which is that playback is easier to debug if neumes are first pre-processed into an intermediate representation. Therefore, I wrote a new analysis service specifically for playback, which transforms the neumes into an intermediate representation. The playback service now deals exclusively with this intermediate representation. I was able to reuse the `getScaleFromFthora` and `getShift` functions from the layout service (now `public` for use in the analysis service), with some minor modifications; however, the ugly `getSpreadIndex` function is still unique to the layout service (its logic is now duplicated in the analysis service) and the concept of valid fthores and fthora restrictions is still exclusively the domain of the layout service.

The biggest improvements to playback are around compound neume spreads. The analysis service is now completely aware of these and breaks them down into note atoms. Fthores, ison markings, and accidentals are applied to the correct note atom in the spread. The "chromatic fthora note" for enharmonic fthores was only used by the playback service for disambiguation purposes, and now there is no ambiguity; therefore, I am removing this concept from the GUI: "chromatic fthora notes" now are exclusively for chromatic fthores.

The idea behind the intermediate representation is that it should be a complete higher-level representation of note durations and pitches. For pitches, this representation consists of a physical note, virtual note (which may be the same as the physical note if no fthora is in effect, or if the fthora happens to match the physical note), and a scale. For durations, this consists of an integer. In addition to note atoms, pitches can be present in mode keys and fthores. These are plumbed through as unique events. Ison and tempo changes are plumbed through as simple events. This all has the effect of normalizing the various complex representations of these abstract concepts into a flat representation that is easy to debug. The first step when examining a playback issue is to examining this intermediate representation and ensure that it makes sense. For example, limitations in the underlying object model such as #318 will show up as problems in the intermediate representation, and it is easy to identify these issues before the playback service begins executing the intermediate representation.

Once the intermediate representation is calculated, it is passed into the playback service for execution. The playback service is a lot easier to read, since it doesn't have logic for interpreting neumes. Two main improvements have been made here: first, frequencies are now calculated based on absolute pitch rather than relative pitch. This approach fixes a number of issues in Makarios Anhr and my Axion Estins, but it is not without its challenges (see below). The logic for enharmonic scale changes has also been significantly simplified. Enharmonic fthores used to terminate at the next martyria, which is incorrect; they are now treated just like any other fthora. I don't attempt to deal with the general fthora termination problems that currently exist in the software in this PR, but that would be the next logical step after this PR. I've also removed the concept of permanent enharmonic Vou: this seems like unnecessary complexity and was never exposed in the UI. In any case, this should not be an issue for any properly notated score and is not a problem for my Axion Estins, which definitely stress test the scale capabilities of the software.

I have played back all my Axion Estins with this PR, and while a handful of them got worse (due to the single remaining major issue described below), almost all of them played back much better than before. Some previously unplayable ones became playable for the first time. And this process has proved invaluable for me as I copyedit the work for final publication, as there are many minor editorial issues that I would not have noticed had it not been for playing them back with (functioning) software. Unfortunately, one major issue still remains, which is why this PR is still in draft state. Recall that when a fthora is encountered, instead of using `moveDistance` and moving relatively based on the current scale, we are now instead jumping to the absolute frequency, calculated based on Di and the target scale determined by the fthora. This works well overall, but there are a few issues:

- For Spathi, calculating the interval based on Di and the target scale determined by the fthora doesn't work because Di isn't stable relative to the Di frequency configured in the playback options. I've hacked around this by special-casing Spathi and doing the calculation relative to Ga or Ke, then incrementing or decrementing by 12 moria to adjust the result to be relative to be Di. I'm open to better solutions here, but this is working pretty well in practice for my two Axion Estins in Hisâr-Bûselik.
- When a fthora is in effect, you need to adjust the result by the number of moria of the transposition (what is called the "shift" in some parts of the code). Computing this isn't trivial. You can't take the number of moria from the physical note to the virtual (transposed) note in the old scale, because that doesn't work when jumping from diatonic Pa to enharmonic Zo. You can't take the number of moria from the physical note to the virtual note in the new scale either, because that doesn't work in other situations. The compromise I settled on, which works a lot of the time, is to compute the distance from the physical note to Di in the old scale, then compute the distance from Di to the virtual note in the new scale, then sum the two. This works whenever the two scales are aligned on Di; however...
- The above solution breaks down if, say, you are in diatonic Ni and put a hard chromatic fthora on Ni or Vou, thus creating a tonal upheaval. The same occurs in the Anoixantaria by Koukouzelis (abbreviated by Chourmouzios) and the end of To tritto tis erotiseos. In these cases, computing the moria of the transposition as described above breaks down, because the original scale and the target scale aren't aligned on Di. As a result the computation is off by an arbitrary amount since the scales don't align with each other, at least not at the point the algorithm is expecting them to. Conceptually, the scales do align, but just not on Di in both scales: in the case of a hard chromatic fthora on diatonic Vou, they align based on physical Zo/virtual Ke, which is intuitive to me as a human but not to the computer. I can't figure out a heuristic to determine this in code.

I will keep thinking about the above problem and trying to come up with a code heuristic for determining the alignment point; however, if a solution cannot be found, we may need frontend support for choosing the alignment point if we want to go this route.

### Testing done

As mentioned above, I have been doing (and continue to do) copious amounts of manual testing with my forthcoming book of over five dozen Axion Estins, many of which are written in nontraditional scales and serve as a good stress test for the scale and playback capabilities of the software. The only regression I am aware of is the tonal upheaval issue mentioned above.

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -4109,11 +4109,6 @@ export default class Editor extends Vue {
       fthora === Fthora.HardChromaticPa_Bottom
     ) {
       chromaticFthoraNote = ScaleNote.Pa;
-    } else if (
-      fthora === Fthora.Enharmonic_Top ||
-      fthora === Fthora.Enharmonic_Bottom
-    ) {
-      chromaticFthoraNote = ScaleNote.Ga;
     }
 
     this.updateNote(element, { fthora, chromaticFthoraNote });
--- a/src/components/ToolbarNeume.vue
+++ b/src/components/ToolbarNeume.vue
@@ -816,20 +816,10 @@ export default class ToolbarNeume extends Vue {
     return [];
   }
 
-  get hasAmbiguousEnharmonicFthora() {
-    return (
-      (this.element.fthora === Fthora.Enharmonic_Top ||
-        this.element.fthora === Fthora.Enharmonic_Bottom) &&
-      this.element.scaleNotes.includes(ScaleNote.Ga) &&
-      this.element.scaleNotes.includes(ScaleNote.Vou)
-    );
-  }
-
   get showChromaticFthoraNote() {
     return (
       this.element.fthora != null &&
-      (this.chromaticFthoras.includes(this.element.fthora) ||
-        this.hasAmbiguousEnharmonicFthora)
+      this.chromaticFthoras.includes(this.element.fthora)
     );
   }
 
--- a/src/models/ModeKeys.ts
+++ b/src/models/ModeKeys.ts
@@ -347,6 +347,7 @@ export const modeKeyTemplates: ModeKeyTemplate[] = [
     scale: Scale.Diatonic,
     scaleNote: ScaleNote.Zo,
     description: 'Enharmonic',
+    fthora: Fthora.Enharmonic_Top,
     martyria: ModeSign.VarysZo,
     note: ModeSign.Zo,
     fthoraAboveNote: Fthora.Enharmonic_Top,
--- a/src/services/LayoutService.ts
+++ b/src/services/LayoutService.ts
@@ -1795,7 +1795,7 @@ export class LayoutService {
     return rootSign;
   }
 
-  private static getScaleFromFthora(fthora: Fthora, currentNote: number) {
+  public static getScaleFromFthora(fthora: Fthora, currentNote: number) {
     if (
       fthora.startsWith('Diatonic') ||
       fthora.startsWith('GeneralSharp') ||
@@ -1853,7 +1853,7 @@ export class LayoutService {
     return null;
   }
 
-  private static getShift(
+  public static getShift(
     currentNote: number,
     currentScale: Scale,
     fthora: Fthora,
@@ -1868,7 +1868,7 @@ export class LayoutService {
       );
 
       shift = fthoraNote - currentNote;
-      shift = shift % 4;
+      shift %= 4;
     } else if (currentScale === Scale.SoftChromatic) {
       const fthoraNote = getScaleNoteValue(
         chromaticFthoraNote ??
@@ -1876,7 +1876,7 @@ export class LayoutService {
       );
 
       shift = fthoraNote - currentNote;
-      shift = shift % 4;
+      shift %= 4;
     } else if (currentScale === Scale.Diatonic) {
       let fthoraNote = currentNote;
 
--- a/src/services/audio/AnalysisService.ts
+++ b/src/services/audio/AnalysisService.ts
@@ -1,0 +1,962 @@
+import {
+  ElementType,
+  MartyriaElement,
+  ModeKeyElement,
+  NoteElement,
+  ScoreElement,
+  TempoElement,
+} from '@/models/Element';
+import {
+  Accidental,
+  Fthora,
+  GorgonNeume,
+  Ison,
+  QuantitativeNeume,
+  TempoSign,
+  TimeNeume,
+} from '@/models/Neumes';
+import { getNeumeValue, getNoteSpread } from '@/models/NeumeValues';
+import {
+  getNoteValue,
+  getScaleNoteFromValue,
+  getScaleNoteValue,
+  Scale,
+  ScaleNote,
+} from '@/models/Scales';
+
+import { LayoutService } from '../LayoutService';
+
+export interface AnalysisNode {
+  nodeType: NodeType;
+  elementIndex: number;
+}
+
+export interface PitchNode extends AnalysisNode {
+  physicalNote: ScaleNote;
+  virtualNote: ScaleNote;
+  scale: Scale;
+}
+
+export interface TemporalNode extends AnalysisNode {
+  duration: number;
+}
+
+export enum NodeType {
+  NoteAtomNode = 'NoteAtomNode',
+  RestNode = 'RestNode',
+  ModeKeyNode = 'ModeKeyNode',
+  FthoraNode = 'FthoraNode',
+  IsonNode = 'IsonNode',
+  TempoNode = 'TempoNode',
+}
+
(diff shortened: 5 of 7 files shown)


src/components/Editor.vue
```diff
@@ -4109,11 +4109,6 @@ export default class Editor extends Vue {
       fthora === Fthora.HardChromaticPa_Bottom
     ) {
       chromaticFthoraNote = ScaleNote.Pa;
-    } else if (
-      fthora === Fthora.Enharmonic_Top ||
-      fthora === Fthora.Enharmonic_Bottom
-    ) {
-      chromaticFthoraNote = ScaleNote.Ga;
```
basil, line 4116: The concept of an enharmonic "chromatic fthora note" was never used except for playback, and there only to disambiguate between multiple options due to lack of awareness of compound neume spreads. The root cause of the issue has now been fixed, so this workaround is no longer needed.

src/components/ToolbarNeume.vue
```diff
@@ -816,20 +816,10 @@ export default class ToolbarNeume extends Vue {
...
 
-  get hasAmbiguousEnharmonicFthora() {
-    return (
-      (this.element.fthora === Fthora.Enharmonic_Top ||
-        this.element.fthora === Fthora.Enharmonic_Bottom) &&
-      this.element.scaleNotes.includes(ScaleNote.Ga) &&
-      this.element.scaleNotes.includes(ScaleNote.Vou)
-    );
-  }
-
```
basil, line 827: Ditto.

src/models/ModeKeys.ts
```diff
@@ -347,6 +347,7 @@ export const modeKeyTemplates: ModeKeyTemplate[] = [
     scale: Scale.Diatonic,
     scaleNote: ScaleNote.Zo,
     description: 'Enharmonic',
+    fthora: Fthora.Enharmonic_Top,
```
basil, line 350: This was previously special-cased. I see no reason not to handle this the same way as other mode keys that have a fthora. As mentioned in the description, I see no need for a "permanent enharmonic Vou" feature. Scores in this mode are rare, and they should be notated properly anyway without the need for software workarounds.

src/services/LayoutService.ts
```diff
@@ -1835,25 +1844,30 @@ export class LayoutService {
     return null;
   }
 
-  private static getShift(
+  public static getShift(
```
basil, line 1847: In order to be able to reuse this from the analysis service, I had to improve it a little. The existing version only ever returns 0 or 1 for chromatic fthores, but I needed it to be able to return 0, 1, 2, or 3 for playback support. Here I improve the accuracy of the result by passing in the chromatic fthora note.

src/services/audio/AnalysisService.ts
```diff
@@ -0,0 +1,967 @@
...
+      // Rewind until we reach the first temporal node or the head node.
+      while (
+        cur > -1 &&
+        nodes[cur].nodeType !== NodeType.NoteAtomNode &&
+        nodes[cur].nodeType !== NodeType.RestNode
+      ) {
+        cur -= 1;
+      }
+
+      // If the piece starts with a gorgon, skip the first duration.
```
basil, line 857: I'm not sure whether this bug exists on trunk or not, but I noticed it while testing, so decided to fix it here.

src/services/audio/AnalysisService.ts
```diff
@@ -0,0 +1,967 @@
...
+
+        // Apply the gorgon.
+        const affectedNode = nodes[cur] as TemporalNode;
+        affectedNode.duration -= durations[i];
+        cur += 1;
+      }
+    }
+  }
+
+  private static normalizeAccidental(accidental: Accidental): Accidental {
```
basil, line 880: Fixes another minor bug I noticed while testing: the playback service wasn't properly handling secondary and tertiary accidentals. It does now, after we normalize them.

src/services/audio/PlaybackService.ts
```diff
@@ -760,958 +526,283 @@ export class PlaybackService {
...
-      modeKeyElement.scale === Scale.Spathi &&
-      modeKeyElement.scaleNote === ScaleNote.Ga
+    if (
+      workspace.lastAlterationMoria !== 0 &&
+      workspace.lastAlterationNote !== noteAtomNode.physicalNote
     ) {
-      workspace.scale = this.spathiGaScale;
-    }
-
-    // TODO support mode keys with enharmonic fthora?
```
basil, line 895: Another benefit of the simplification in this PR is that this TODO now gets done for free.

src/services/audio/PlaybackService.ts
```diff
@@ -760,958 +526,283 @@ export class PlaybackService {
...
+            fthoraNode.virtualNote +
+            ' to Di in the new scale',
+          moria2,
+        );
+      }
 
-    if (noteElement.timeNeume != null) {
-      oligonDuration +=
-        this.timeMap.get(noteElement.timeNeume)! * workspace.beat;
+      workspace.transpositionMoria = moria + moria2;
```
basil, line 698: Note: this is the logic that works "most" of the time, but not when there is a tonal upheaval, hence this PR being in draft state.

basil:
I cleaned up the spathi code a little but I am still no closer to figuring out how to handle tonal upheavals correctly, or even a sensible parameterization of the function to compute the transposition moria.

basil:
From my perspective this PR is ready.

danielgarthur (approved):

PR #343, opened 2023-10-16
Barline support could use some work

### LINKED ISSUE #151: Barline support could use some work
Try typesetting e.g. https://digitalchantstand.goarch.org/goa/dcs/js/viewer/web/viewer.html?file=/m/jmboyer/gr/ho/s03/ma/b/resevlogetaria.pdf in Neanes and you will see that it is not currently possible. Apart from the issues already raised in #144, the following additional issues exist:

- Suppose you generally attach a barline to the beginning of the first note of a measure. If that measure is preceded by a martyria, some style conventions move the barline to be centered above the martyria. Neanes has no easy way to achieve this without tedious manual positioning.
- Suppose you generally attach a barline to the beginning of the first note of a measure. If that measure begins a new line, you generally also want the end of the previous line to contain a barline, as in lines 4 and 5 of the linked score. In Neanes you could achieve that by manually putting a barline at the end of the previous line, but that is a workaround rather than a real solution because the piece could reflow with a different line break and the workaround would fall apart.

basil:
Depends on https://github.com/danielgarthur/sbmufl/pull/11. Fixes #151.


src/components/Editor.vue
```diff
@@ -3865,12 +3866,15 @@ export default class Editor extends Vue {
     // Right
     // Both Sides
     // None
-    if (neume === element.measureBarLeft && neume === element.measureBarRight) {
+    const normalizedMeasureBar = element.measureBarLeft?.endsWith('Above')
```
basil, line 3869: UI remains unchanged, the only difference being that applying a "left" barline to a multi-neume group now automatically converts it to a top barline in `Element.ts` (same for previously serialized scores), and apply a "left" barline to a martyria now automatically converts it to a top barline. Here we translate the actual neume into the simplified UI concept, and in `Element.ts` we ensure that the simplified concept (or old scores) are appropriately converted.

src/services/integration/ByzHtmlExporter.ts
```diff
@@ -1273,6 +1273,12 @@ export class ByzHtmlExporter {
     map.set('barlineShortSingle', 'x-bar-s');
     map.set('barlineShortDouble', 'x-bar2-s');
     map.set('barlineShortTheseos', 'x-bar-th-s');
+    map.set('barlineSingleAbove', 'x-bar');
+    map.set('barlineDoubleAbove', 'x-bar2');
+    map.set('barlineTheseosAbove', 'x-bar-th');
+    map.set('barlineShortSingleAbove', 'x-bar-s');
+    map.set('barlineShortDoubleAbove', 'x-bar2-s');
+    map.set('barlineShortTheseosAbove', 'x-bar-th-s');
```
basil, line 1281: I don't use ByzHtml, so I don't know what needs to be done here. I just did the minimal change to get the tests to stop complaining.

PR #470, opened 2023-12-14
Add a localization framework

basil:
Start the process of localizing this application by introducing a localization framework and converting about half of the application to use it. I chose `i18next` because it was lightweight and worked fine in both Electron and Vue, whereas I had trouble getting other frameworks like `vue-i18n` to work on the Electron side. On the Electron side we rely on `app.getLocale()`, which on my Linux system is set to `en-US` by default and `el` when I set `LANGUAGE=el` before starting Electron. On the Vue side we rely on either `navigator.languages` (which Electron sets the same way as `app.getLocale()`, if running in Electron; otherwise, the user's browser language preferences) or the `?lng` query string parameter for ease of testing.

The software has lots of strings that need to be pulled out into translation files: menus, toolbars, dialog boxes, and tooltips. I got through the menus and toolbars before I had enough for one day. The rest can be done in subsequent PRs, but I wanted to post this initial PR in case there was negative feedback about the general approach before I spend too much time on this. I freely admit I have no idea what I am doing when it comes to JavaScript development, so if I am doing something wrong, then please tell me and I will try my best to figure it out.

As a proof of concept, I localized a handful of strings in both Greek and Romanian. You can test this out in the web app by setting the query string `?lng=el` or `?lng=ro`, or by starting the Electron app after setting the environment variable `LANGUAGE=el` or `LANGUAGE=ro` (at least on Linux; I have no idea how this works on other platforms). I don't actually speak these languages, so the hope is that users will contribute more localizations. Once I finish converting all the dialog boxes and tooltips I might try to do an automated translation with Google Translate and/or ChatGPT, even if the results are awkward. Nothing gets people to contribute like fixing an awkward translation!

To test that the `en.json` file was complete and not introducing any regressions, I used https://www.npmjs.com/package/pseudo-localization to mutate the English strings to accented variants and verified that the app looked identical modulo the accent marks.


danielgarthur:
This looks good so far.

basil:
Great! I localized the dialog boxes. I have not (yet) bothered to localize the tooltips. Much remains to be improved, but localization is a never-ending battle and this seems as good a place to start as any. Further improvements can be made in subsequent PRs.

danielgarthur (approved):

PR #473, opened 2023-12-16
Add pseudolocalization support

basil:
Make it easier for others to test localization by formally integrating a [pseudolocalization](https://en.wikipedia.org/wiki/Pseudolocalization) framework and adding instructions to the documentation.


efraimkaov:
I would be  happy to help with translation into Romanian and maybe Greek if no one else is helping (right now I'm busy with some projects). Just let me know when everything is set up and ready to start up with the translation.

basil:
@efraimkaov This is an open source project, so _you_ are the person who is helping. :smile: Feel free to start working on the `el.json` and `ro.json` files once #474 is merged.

electron/main/index.ts
```diff
@@ -1612,11 +1613,20 @@ app.on('activate', () => {
...
 app.on('ready', async () => {
-  i18next.init({
-    lng: app.getLocale().slice(0, 2),
-    fallbackLng: 'en',
-    resources,
-  });
+  i18next
+    .use(
+      new Pseudo({
+        enabled: 'VITE_PSEUDOLOCALIZATION' in import.meta.env,
```
danielgarthur, line 1619: Should this instead be `import.meta.env.VITE_PSEUDOLOCALIZATION`? That way setting to false won't enable it.

danielgarthur: The new property can be added to `src/env.d.ts` to appease Typescript.

basil: > Should this instead be `import.meta.env.VITE_PSEUDOLOCALIZATION`?

I do not think so, because if it is, then pseudolocalization is enabled even when no `.env.local` file exists.

danielgarthur: All env variables are treated as strings, so you would have to cast it to a boolean or check that it equals the string "true". As it is written now, I would think it would enable pseudolocalization regardless of what value the variable is set to, since it is only checking for the existence of a variable with that name. I can test it tomorrow, but I think this is the case.

danielgarthur (approved):

PR #579, opened 2024-03-18
Fix measure bar left applied wrongly to previous line.

ilizol:


basil:
Not sure if this will cause any regressions for my scores at https://github.com/basil/byzantine-music. In general I think barline support needs some more work and UI improvements to deal with this case as well as a few others I discovered in the above scores. I will try to find some time to work on it in the next few weeks.

basil:
Causes #665

ilizol:
Add #151 as a reference.

danielgarthur:
Talking with @ilizol, it sounds like the issue is not that there is a bug in the bar line logic, but that rather he would rather the "courtesy bar line" not be added. In the same way Musescore has an option to disable courtesy key signatures, Neanes should allow this courtesy bar line to be disabled.

basil:
That makes sense to me and would definitely be an improvement over the status quo, but I suspect a fully-fledged UI would still have to cover some additional cases. For example, look at lines 2-3 of https://cappellaromana.org/wp-content/uploads/2023/10/Sun-of-Justice-Sample-Page_10-5.jpg and lines 1-2 of https://cappellaromana.org/wp-content/uploads/2023/10/Sun-of-Justice-Sample-Page_10-4.jpg. There the "courtesy bar line" is different from the barline on the next/previous line. My existing inference logic doesn't support this, and some sort of UI support would be needed, but the UI support would need to be more complex than the checkbox described above.

Issue #615, opened 2024-04-11
Lyric import/export


basil:
I am editing my second book with Neanes, a 75-page work of the music for the Small and Great Supplicatory Canons. Unlike the first collection of Axion Estins, this work is very text-heavy with dozens of pages of heirmologic troparia. Without native spell-checking or hyphenation support, it is difficult to achieve high-quality output if your only option for lyric entry is to enter lyrics syllable by syllable into the score. Even starting from a document with (software-produced) perfect hyphenation, I made [a large number of typographical errors](https://github.com/basil/byzantine-music/commit/e3f5c070f29787dad3c64c505aad0ac242f7edc3) while entering the lyrics into the score, and the only way for me to discover them was to write a Python script to export lyrics out of Neanes and compare them to my input. I don't imagine this is a practical option for most users.

Both Finale and NoteWorthy Composer offer a superior solution in the form of a dedicated lyrics editor. Both offer the ability to edit and export existing lyrics. The version in Finale allows new lyrics to be entered into the document, but they must be placed onto notes with mouse clicks. [The version in NoteWorthy Composer](https://noteworthycomposer.com/faq/15.md) is somewhat superior, as it allows new lyrics to be entered in the lyrics editor and they are automatically placed onto notes based on the number of hyphens, underscores, or spaces present in the lyrics editor dialog. This allows for efficient import _and_ export of lyrics and also allows lyrics to be easily transferred from one document to another.

I would very much like for a NoteWorthy Composer style lyrics editor to be added. This way, you can type lyrics in your word processor of choice, spell check them there, run them through e.g. https://juiciobrennan.com/hyphenator/ to hyphenate them correctly, and paste the result into the Neanes lyric editor for efficient import, adding extra hyphens, underscores, or spaces for lyrics that span more than one syllable as needed. This also enables the opposite use case, where lyrics can be entered interactively in Neanes and then exported to a word processor for spell checking by opening the lyrics editor, copying the lyrics, pasting them into the word processor, removing the hyphens, and running the spell checker. Removing the hyphens is a bit of a pain and doesn't work 100% of the time because "all-ho-ly" would become the incorrectly spelled "allholy", so it is preferable to do the spell checking before the hyphenation, but the latter workflow is still better than nothing when you have a document with lyrics you don't trust already entered.

[...]

PR #633, opened 2024-04-17
Advanced Lyrics Entry

danielgarthur:
This PR adds a new way of entering lyrics that makes it easier to create prosomoia, as well as to quickly shift lyrics. The file `advanced.md` in this PR has a more detailed description of the feature.

It does not add automatic hyphenation or spell-checking.

#615


basil:
Drop cap handling doesn't seem correct for paragraphs that begin with a single-letter word, like "I" or "O"—in these cases, the drop cap should include a space, as HTM's publications do. See https://github.com/basil/byzantine-music/blob/27955c0002750c7004c2c3a78632536c92f66221/Mysteries%20and%20Sacraments/Paraklesis/24%20Megalynaria.byzx.

PR #642, opened 2024-04-20
Add Stathis Series font

danielgarthur:
This PR adds the Stathis Series font (LTR only). Font may be selected in the Page Setup dialog.


basil:
Causes #665

PR #659, opened 2024-04-23
Handle whole-word drop caps

basil:
https://github.com/neanes/neanes/pull/633#issuecomment-2068121485

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -4738,12 +4738,14 @@ export default class Editor extends Vue {
       if (filteredElements[i].elementType === ElementType.DropCap) {
         const dropCap = filteredElements[i] as DropCapElement;
         const token = tokenizer.getNextCharacter();
+        const nextToken = tokenizer.peekNextToken();
+        const content = nextToken === '_' ? `${token} ` : token;
 
         if (dropCap.content !== token) {
           updateCommands.push(
             this.dropCapCommandFactory.create('update-properties', {
               target: dropCap,
-              newValues: { content: token },
+              newValues: { content },
             }),
           );
         }
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -4741,7 +4741,7 @@ export default class Editor extends Vue {
         const nextToken = tokenizer.peekNextToken();
         const content = nextToken === '_' ? `${token} ` : token;
 
-        if (dropCap.content !== token) {
+        if (dropCap.content !== content) {
           updateCommands.push(
             this.dropCapCommandFactory.create('update-properties', {
               target: dropCap,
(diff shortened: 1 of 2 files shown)


src/components/Editor.vue
```diff
@@ -4738,12 +4738,14 @@ export default class Editor extends Vue {
       if (filteredElements[i].elementType === ElementType.DropCap) {
         const dropCap = filteredElements[i] as DropCapElement;
         const token = tokenizer.getNextCharacter();
+        const nextToken = tokenizer.peekNextToken();
+        const content = nextToken === '_' ? `${token} ` : token;
 
         if (dropCap.content !== token) {
```
danielgarthur, line 4744:
```suggestion
        if (dropCap.content !== content) {
```

danielgarthur (changes requested):
Could you give a visual example of what you are try to do (i.e. a picture from a published book)? When I try the code, it results in a rather large gap.
image
I'm familiar with drop caps having some extra space in printed text when it's a one letter word, but I'm not sure I've seen neume lyrics with such large gaps.

danielgarthur:
Sorry, I just remember you had a link to your file. I see what you are going for. Should this gap always be desired or should it be a setting to turn on and off?

danielgarthur:
Or perhaps `O_` could result in the extra space, but `O _` could not?

basil:
I checked my copy of Robert Bringhurst's _The Elements of Typographic Style_, and he writes the following:

> In English, if the initial letter is A, I, or O, a question can arise: is the initial letter itself a word? The answer to this question must come in the spacing of the text in relation to the versal. If the first word of the text is _Ahead,_ for example, excessive space between the initial A and the rest of the word is bound to cause confusion.

> I'm familiar with drop caps having some extra space in printed text when it's a one letter word, but I'm not sure I've seen neume lyrics with such large gaps.

That might just be because nobody has yet produced an English book of Byzantine music with this level of typographical rigor. I do not see a reason not to make this the only supported behavior.

danielgarthur:
That seems logical and I am an agreement that this should be the only supported behavior. However, with the PR how it currently is, it seems like there is no way to distinguish between `Ahead` and `A head` in the lyrics. No matter what I do, I always get the extra space.

basil:
I think the existing code fails to hyphenate a two-syllable word like "a-head" where the first syllable is a drop cap—even when typing a hyphen on the first syllable, no hyphen appears after typing the second syllable. I'm not making that problem any worse than it is already.

If you think this PR is introducing a bug, then please give me detailed steps to reproduce it from scratch.

danielgarthur:
Oh, right... I need to get some sleep because I completely forgot that I should type `A-head` to get the results I want for `Ahead`.

danielgarthur (approved):

PR #660, opened 2024-04-23
Look ahead past martyriae, drop caps, and mode keys

basil:
Even after fixing #659, my Dormition Exaposteilaria weren't round-tripping properly through the lyrics tokenizer. I tracked down the problem to a martyria, which was stopping the extraction process from adding the last underscore. Fixed by looking ahead past martyriae. With this change and #659 my Dormition Exaposteilaria round-trips correctly.

### DIFF
--- a/src/services/LyricService.ts
+++ b/src/services/LyricService.ts
@@ -65,14 +65,16 @@ export class LyricService {
             lyrics += '-';
           }
         } else if (note.isMelisma) {
-          const nextNote =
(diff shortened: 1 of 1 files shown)


danielgarthur (approved):

basil:
Testing of the Post-Gospel Troparia from my Paraklesis book showed that we also need to look ahead past drop caps and mode keys as well.

PR #669, opened 2024-04-27
Fix lyrics look-ahead logic

danielgarthur:
Amends #660, which appears to be missing a `break;` statement. This caused the `nextNote` to be always the last note.
After:
image

Before:
image


basil:
Thank you Daniel and sorry for the trouble. Although I did test that other PR against every Neanes score I have, I didn't have any scores that triggered this specific failure mode. I guess that underscores the need to develop an open-source test suite for this software. I have been trying to do my part to develop such a thing over at https://github.com/basil/byzantine-music.

PR #709, opened 2024-05-23
Fix number of underscores when MelismaOnly is set

danielgarthur:


danielgarthur:
Oops. Glad I wrote those unit tests. 

This needs a smarter solution. A note that starts a melisma will have to look ahead to see whether it has any non-`MelismaOnly` notes in the melisma. If so, it should add the extra underscore. Otherwise it should not. 

I'll fix this soon.

Issue #640, opened 2024-04-18
Rich text editor


panosmorningstar:
Hi. 

Neanes is truly remarkable, very easy to use with a clean UI.

I propose to the developers the creation of different kinds of text boxes. This will result in the enrichment of the document with plain lyrics, creating whole ἀκολουθίαι. This could also include specific settings for the footnote/header text boxes.

Thank you.

danielgarthur:
@panosmorningstar Would you please upload some pictures or PDFs that illustrate the kind of text you would like to see supported?

panosmorningstar:
@danielgarthur 
[Uploading 2024-04-28_kyriakh_baiwn_e_(scores).pdf…]()

danielgarthur:
The upload did not seem to work. Could you please try again?

danielgarthur:
Never mind, I was able to find the PDF on the internet.

[2024-04-28_kyriakh_baiwn_e_(scores).pdf](https://github.com/neanes/neanes/files/15049251/2024-04-28_kyriakh_baiwn_e_.scores.pdf)

danielgarthur:
Based on the file, it seems like the main thing is that Neanes does not currently rich text boxes that allow different font colors and styles in the same box. 

This is something that I've looked into several times in the past and the conclusion that I've come to every time is that it would be best to write my own rich text box, as opposed to trying to use an existing library.

I've put this off for a while to focus on other more essential features, but it may be time for me to look into rich text boxes again.

basil:
Over a decade ago I enjoyed using [TinyMCE](https://github.com/tinymce/tinymce), but I have no idea if it is still relevant these days.

danielgarthur:
Added to release [0.4.32](https://github.com/neanes/neanes/releases/tag/v0.4.32).

There is an example file in the examples folder [here](https://github.com/neanes/neanes/blob/master/examples/Rich%20Text.byzx) and a PDF of the example [here](https://github.com/neanes/neanes/blob/master/examples/Rich%20Text.pdf).

danielgarthur:
Re-opening because the original request also includes rich text support in headers and footers. This is not completed yet, but I think it's worth including.

danielgarthur:
Headers and footers have been added to [v0.4.34](https://github.com/neanes/neanes/releases/tag/v0.4.34).

PR #716, opened 2024-05-28
Rich text editing

danielgarthur:
#640

This PR adds a rich text box that uses [CKEditor](https://ckeditor.com/docs/ckeditor5/latest/installation/integrations/vuejs-v3.html). The HTML from the editor is extracted and saved in the `byz` file. This keeps the path open for allowing users to insert custom HTML and to swap out with any other HTML editor in the future.

A shortcoming is that the text boxes will not split across to the next page if you reach the end of a page. Previously, this was not much of an issue because text was meant to be minimal. But including rich text may encourage users to attempt to build entire services, which might include entire psalms and litanies. This means users must currently split the content themselves among multiple text boxes.

This issue will not be addressed in the current PR and I'm not sure of a good way to address it. We can detect the height of text boxes. This could be used to create or merge textboxes, but I'm not sure how to do that in a smooth way (i.e. seamlessly as the user types).


src/components/TextBoxRich.vue
```diff
@@ -0,0 +1,225 @@
...
+      :style="textBoxStyle"
+      class="rich-text-editor"
+    ></ckeditor>
+  </div>
+</template>
+
+<script lang="ts">
+import { EditorConfig } from '@ckeditor/ckeditor5-core';
+import { FontSizeOption } from '@ckeditor/ckeditor5-font/src/fontconfig';
+import { CKEditorComponentData } from '@ckeditor/ckeditor5-vue/dist/ckeditor';
```
basil, line 24: Removed in https://github.com/ckeditor/ckeditor5-vue/pull/281.

Issue #812, opened 2024-07-05
Line margins changing back and forth while typing


pabaxe:
**This bug was found in**

- [x] Desktop
- [ ] Web App
- [ ] Both

**Describe the bug**
While writing a score, after a certain line was filled and the cursor went automatically to the next line, the margins of the line above change each time something is typed, and then changes back and forth continuously. I'm attaching a video to better describe it. 

In the video, all I'm doing is pressing (J) and deleting, to showcase the issue. 

**Screenshots or files**

https://github.com/neanes/neanes/assets/174416975/cd62c9e8-133f-4015-a36a-7efc15b905e6

**Desktop Version (please complete the following information, if applicable):**

- OS: MacOS Sonoma 14.5
- Version 0.4.45

danielgarthur:
@pabaxe Could you attach the file? You'll have to put it in a `zip` archive or Github won't let you upload it.

@basil The only thing that I can think of from the video is that the logic that automatically adds bar lines to the last neume in the previous line is adding the bar line, but that pushes the width of the line over the edge, so it breaks to a new line, and now no longer needs the bar, so now it can fit on the previous line again. And so the cycle goes back and forth.

basil:
Yeah, I've seen this problem in a lot of my own scores as well, and I've long suspected that what you described above is the root cause.

pabaxe:
Yes after messing around I found that the only way to bypass it is by removing the bar lines. Sometimes, if the bar line is double, making it single also works, but not always.

danielgarthur:
@pabaxe I've found a way to reliably reproduce this, so your file is not necessary.

danielgarthur:
Fixed in [0.4.46](https://github.com/neanes/neanes/releases/tag/v0.4.46)

PR #814, opened 2024-07-05
Fix measure bar jitters

danielgarthur:
This PR uses the "additional width" feature in the layout processor to account for the additional width of a measure bar that will be added due to the presence of a measure bar at the beginning of the next line. This prevents the jitters seen in #812.

It also fixes another issue where notes where considered tied to other notes or martyria even when a page or line break was present.


basil (approved):
This is great! Sorry for introducing this issue and thank you for the PR.

Issue #213, opened 2023-08-12
Diatonic Zo attraction algorithm is too simplistic


basil:
The algorithm for Diatonic Zo attraction in the playback service is too simplistic. This is pure Agia, and as such if the melody descends past Zo to Ke and then ascends back to Zo, the Zo remains natural and the Ke is attracted (sharp) to Zo. That is, if the melody is revolving around Zo, the Zo remains natural even if it temporarily descends to Ke. The Zo is only flattened if it descends past Ke to Di (in which case Zo is flat and Ke is natural). This illustrates the problem:

![agia-attractions-1](https://github.com/danielgarthur/neanes/assets/29850/9db283ae-04fd-4576-8e1c-250c347573d5)

basil:
Algorithm should be something like this:

* Suppose you are not on Zo and encounter a Zo. This Zo is the input to the function, called _pivot_. Keep scanning ahead until you find the next non-Zo note _next_.
  * If _next_ is above Zo, then Zo _pivot_ (and any other Zos between _pivot_ and _next_, if they exist) remains natural.
  * If _next_ is below Ke, then Zo _pivot_ (and any other Zos between _pivot_ and _next_, if they exist) is flattened.
  * If _next_ is Ke, keep scanning until you find the next non-Ke note _next′_.
    * If _next′_ is above Ke, then Zo _pivot_ (and any other Zos between _pivot_ and _next_, if they exist) is natural and Ke _next_ (and any other Kes between _next_ and _next′_, if they exist) are sharp. If _next′_ is a Zo, the function recurs with _next′_ being passed in as _pivot_.
    * If _next′_ is below Ke, then Zo _pivot_ (and any other Zos between _pivot_ and _next_, if they exist) is flat and Ke _next_ (and any other Kes between _next_ and _next'_, if they exist′) is natural.

basil:
I can implement this once #322 goes in.

danielgarthur:
Fixed in [0.4.44](https://github.com/neanes/neanes/releases/tag/v0.4.44).

Issue #821, opened 2024-07-09
Playback of Ke in Third Mode. 


PeTH2:
Linux, version 0.4.46.

In certain theseis that end on Ke, especially when Ke is followed by Zo, Ke plays back as flat Zo.

Here's a screen recording:

https://github.com/neanes/neanes/assets/128826899/dbcb8ccf-741e-4689-9160-096678729f46

danielgarthur:
Related to #213. Unfortunately the Ke attraction rules described in that issue are also too simplistic. I'm going to disable Ke attraction for now until a more complete algorithm can be determined.

PeTH2:
Thank you!

danielgarthur:
Fixed in [0.4.47](https://github.com/neanes/neanes/releases/tag/v0.4.47). 

#606 should still be working, despite the partial rollback of the zo attraction feature.

PR #823, opened 2024-07-09
Disable ke attraction

danielgarthur:
This PR disables the ke attraction algorithm from #213.

The algorithm does not currently distinguish between agia and non-agia melodies such as third mode melodies centered around ke.


danielgarthur:
#821

PR #836, opened 2024-07-13
Display an error message when invalid zoom is selected

basil:
Clearer to show an error rather than to silently discard the input in my opinion.

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -1735,12 +1735,6 @@ export default class Editor extends Vue {
   }
 
   set zoom(zoom: number) {
-    if (zoom < 0.5) {
-      zoom = 0.5;
-    } else if (zoom > 5) {
-      zoom = 5;
-    }
-
     this.selectedWorkspace.zoom = zoom;
   }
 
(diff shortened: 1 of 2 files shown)


danielgarthur:
If we do show an error, I'd rather use something less intrusive than `alert`. Either a toast/snackbar message, or a little notification that appears under the input (like a form validation speech bubble).

Alternatively, we could expand the zoom to whatever you think is reasonable such that reversion are unlikely. I think MS Word does 10-500%. LibreOffice Writer is 40(?)-600%. Maybe we could do 10% - 1000%? There's no real barrier to what we choose.

basil:
> If we do show an error, I'd rather use something less intrusive than `alert`.

What is the reason for this request? I checked my copies of Word 2000 and InDesign CS 2, and they show an intrusive alert just like I am doing above.

danielgarthur:
The reason is that I'm used to web development and web design generally discourages `alert`. But this is a desktop app and I see that the latest version of Word is still showing an alert, so I'm okay with the alert. What does the alert look like though? Showing the alert via the Electron's main process may look better.

danielgarthur:
For example, like this code from `Editor.ts`

```
if (error instanceof Error) {
        if (this.ipcService.isShowMessageBoxSupported()) {
          this.ipcService.showMessageBox({
            type: 'error',
            title: 'Open failed',
            message: error.message,
          });
        } else {
          alert(error.message);
        }
      }
```

basil:
I have updated this to use `showMessageBox` when available, as well as to reject invalid values rather than silently rounding them to the closest valid value.

danielgarthur (approved):

basil:
> If we do show an error, I'd rather use something less intrusive than `alert`. Either a toast/snackbar message, or a little notification that appears under the input (like a form validation speech bubble).

→ #2215

PR #842, opened 2024-07-14
Alternate glyphs for small or petite capitals

basil:
Initial implementation is only for (non-inline) text boxes, as I couldn't figure out how to get the width calculation right for inline text boxes. I suppose we'll need to write a custom CKEditor5 plugin to do this for rich text boxes. I'd be happy to work on that, but I'm not sure how you would want to integrate such a plugin into our build system.


basil:
This works, but I wonder if it might be better in the long term to avoid cluttering the text box types and focus all energy on a CKEditor 5 plugin for advanced typography. I started writing one today and, while tedious, it doesn't seem too difficult.

danielgarthur:
The advantage of the plug-in would be that you could assign different feature settings to different paragraphs (or maybe even spans?).

The cheap way to do it would be to instead assign CSS classes or styles to the whole editor for each checked feature. But then it always applies to the whole textbox.

Regarding regular text box sizes, if these font features were used, we would most likely need to adopt the pattern used by rich text boxes. That is, the text box should let its size be determined by Chromium, and notify the `Editor` whenever its size changes. The complication here is that if a text box uses the default styling, then whenever the styling is changed in Page Setup, we have to re-calculate the size of every text box. This is being done in a hacky way that involves rendering all the text boxes off screen so that Chromium can figure out their sizes.

If we want to go down that route, I can help implement that portion of it. Since `TextMetrics` does not support font-features as far as I can tell, even the height of non-inline text boxes could end up being incorrect unless something like this is implemented.

basil:
Yesterday I hacked together a working small caps button for CKEditor 5. It was very tedious, but not intellectually challenging, mostly due to their excellent developer documentation and https://github.com/ckeditor/ckeditor5-package-generator. I will try to add more features to it this week to turn it into a `ckeditor5-typography` package with support for all the things I've been looking for. At that point we can try it out in Neanes.

In the long term, I wonder if we should deprecate for removal the old-style text boxes and commit to CKEditor 5 for everything. Having two implementations in the software seems confusing to end users as well as difficult to maintain.

PR #888, opened 2024-08-01
add the fthora note in the diatonic modes and third mode

cryoranger:
In recent publications and also some theory books the fthora in the mode key is not omitted.
In the mode keys of chromatic modes as seen in neanes, the fthora is not omitted.
Based on the printouts I had, I have added the appropriate fthoras to the mode keys of the diatonic modes and the third mode.


danielgarthur:
I don't have any issue with the fthora being used, but I would like to give users the option since diatonic is the default and older books don't include it.

While there are more complicated options we could implement, I think for now I'm fine with just adding these as duplicates and users can choose between the option with or without the fthora.

More complicated options include either implementing a checkbox filter in the mode key dialog that filters out the duplicates based on the selection, or alternatively introducing an "optional fthora" property for mode keys and the checkbox determines whether the optional fthora is applied.

PR #916, opened 2024-08-19
Option to Disable Automatic Greek Melismata

pathaddad:
This feature is useful whenever Greek letters are used in conjunction with another language, for example for Arabic phonetics:

image

Added as a checkbox option in the "Miscellaneous" section of the Page Setup

image


danielgarthur (changes requested):
Thanks for the PR. A couple things to note:

- The new property on PageSetup needs to be added also to the PageSetup model in `models/save/v1` and the `SaveService` needs to be updated to save and load the new field.
- I think all the instances where `MelismaHelperGreek.isGreek` is called will need to be updated to check this variable. For example, in the `LyricService`.
- I'm thinking the variable name should be something more specific like `disableGreekMelismata`.

danielgarthur (approved):

pathaddad:
Thank you for completing this @danielgarthur

PR #933, opened 2024-08-29
Bump `vue-tsc` to 2.1.0

basil:
Less aggressive than #928. Will it fail with the same error?


basil (commented):
https://github.com/vuejs/language-tools/issues/4748

PR #1067, opened 2024-10-26
Add MusicXML export

danielgarthur:
This PR contains a first pass at exporting to XML.

Future improvements:

- Greek lyrics do not render correctly
- Accents should be added on certain petaste and vareia notes
- Measure calculation needs to be improved when no bar lines, or minimal bar lines are present.
- Add support for compressed MusicXML files ([reference](https://www.w3.org/2021/06/musicxml40/tutorial/compressed-mxl-files/))
- Add support for Segah legetos


src/services/integration/MusicXmlExporter.ts
```diff
@@ -0,0 +1,1454 @@
...
+
+  getDuration(node: TemporalNode, workspace: MusicXmlExporterWorkspace) {
+    const rounded = node.duration.toFixed(2);
+
+    let duration = node.duration;
+
+    // Digorgon is interpreted as 1/2 + 1/4 + 1/4
+    // This is primarily because it is easier for singers to sight sing,
+    // and is a convention used in St Anthony's Divine Music Project
+    // TODO support true triplets as a configurable option
```
basil, line 895: @danielgarthur I'm doubting whether this judgment is accurate. Can you provide examples? Byzantine notation sometimes uses a trigorgon to notate what in practice is often sung as a dotted trigorgon. I would call these "implied dotted trigorgon" and in my scores I try to be explicit about what the classical scores merely imply. In other cases the trigorgon should really be chanted as a triplet (for example in ornamented pieces like the beginning of By the Waters of Babylon). I doubt that this was done in St Anthony's Divine Music project to make the music easier for singers to sight sing, rather I think you might have been thinking about cases of the former. The problem is this code then makes transcriptions of the latter incorrect.

danielgarthur: The Dogmatic Theotokion of the second mode at "arisen".

danielgarthur: First mode Lord I have cried, too. I thought somewhere it is explained why the choice was made although I didn't find it skimming through the Vespers introduction. The explanation was that sight singers often aren't skilled at chanting triplets and that many chanters chant it that way anyway. Not saying it's correct but it's a reality.

danielgarthur: I not saying that this should remain as is either. But as you suggested elsewhere, some kind of plugin or configuration pattern is necessary to make everyone happy and allow different levels of analysis and complexity.

basil: @danielgarthur I would sing both of those two examples with an eighth and two sixteenths (as opposed to a triplet), as they are transcribed. And I would do so deliberately, not because I'm not skilled at chanting triplets. That is just a very common analysis of those phrases. In classical scores the dot is merely implied, but in many analytical books (also written by the kind of person who would have no problem with chanting a triplet) the dot is added. But there are other places where a trigorgon is not analyzed that way and should always be chanted as a triplet. So the mistake in this PR is not in how those two examples are transcribed per se, but rather that this is done (erroneously) unconditionally rather than for the specific formulae for which that analysis applies (and the fact that the comment above the code gives the wrong reason). So the default code path should produce triplets, and then when #1624 is implemented the (hypothetical future) basic analysis module should detect the cases in which the dot is implied and analyze only those cases appropriately as an eighth and two sixteenths.

PR #1235, opened 2025-01-26
Fix page margin bug (extra margin on printing to PDF).

ilizol:


danielgarthur:
What issue does this fix? What are the reproduction steps to see the issue so that I can verify this fixes it?

ilizol:
> What issue does this fix? What are the reproduction steps to see the issue so that I can verify this fixes it?

While I was working with the [Almouzios](https://github.com/ilizol/Almouzios) font, my test files were the followings:

- https://github.com/ilizol/neumes/blob/main/kontakion_prophet_elias_2.pdf
- https://github.com/ilizol/neanestex/blob/add-almouzios-font/examples/kontakion_prophet_elias_almouzios_latex.pdf
- https://github.com/ilizol/neanes-examples/blob/main/kontakion_prophet_elias/kontakion_prophet_elias_almouzios.pdf

Comparing always these files while making changes to the font, found that not only the pdf A4 page size is wrong from Neanes export, but also there is a margin at the left side of the page in the pdf file that was exported from Neanes.

The issue can be easily reproduced if the title is left aligned with zero page margins.

ilizol:
The margin, as I remember, is 1 pixel.
Maybe the guidelines cause this margin, either they are visible / enabled or not?

PR #1269, opened 2025-02-15
Fix some measure bar issues in ByzHtmlExporter

dani31x:
I'm trying to fix the following bugs:

- [x] vareias are added after measure bars in the exported .html
- [x] the exporter doesn't support short measure bars
- [ ] the exporter doesn't support measure bars above martyria (I don't think that I can solve this, actually)

### DIFF
--- a/src/services/integration/ByzHtmlExporter.ts
+++ b/src/services/integration/ByzHtmlExporter.ts
@@ -478,10 +478,17 @@ export class ByzHtmlExporter {
     let inner = '';
 
     if (element.measureBarLeft) {
-      inner += this.exportNeume(MeasureBar.MeasureBarRight, indentation + 2, {
-        x: element.measureBarRightOffsetX,
-        y: element.measureBarRightOffsetY,
-      });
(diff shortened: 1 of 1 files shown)


src/services/integration/ByzHtmlExporter.ts
```diff
@@ -477,20 +477,20 @@ export class ByzHtmlExporter {
   exportNote(element: NoteElement, pageSetup: PageSetup, indentation: number) {
     let inner = '';
 
+    if (element.measureBarLeft) {
+      inner += this.exportNeume(element.measureBarLeft, indentation + 2, {
+        x: element.measureBarRightOffsetX,
+        y: element.measureBarRightOffsetY,
```
dani31x, line 483: What are those, actually? Are they "right offsets" as in "offset on the right", or as in "offsets of right measure bars"?

danielgarthur: The offsets should be changed to `Left`

danielgarthur (approved):

PR #1270, opened 2025-02-15
Export --byz-neume-font-family

dani31x:
Part of https://github.com/danielgarthur/byzhtml/issues/2


danielgarthur:
The special case for melkite RTL where it sets the default font family, is no longer needed, I don't think.

danielgarthur (approved):

Issue #1017, opened 2024-10-07
Stavros, pause and breath mark


gugushvili:
This is a minor bug: it is possible to enter lyrics below stavros, pause sign, or breath mark (upper comma).

The behaviour should be like currently with martyrias, which are automatically left out in the lyrics entry process.

PR #1307, opened 2025-03-05
prevent navigation to lyric boxes for no-lyric note elements

cmillr:
Hi, this is my attempt to fix the issue reported at #1017

I couldn't reproduce this issue using the lyrics toolbar; I observed the issue when entering lyrics directly in the lyrics container beneath the neume.

This fix should skip over the lyrics containers beneath the cross, apostrophe and dotted vareia while entering lyrics in that way, when navigating using the left/right arrow keys, spacebar, or hyphen/underscore melisma.  It is still possible to select the lyrics container on these neumes with the mouse, or with the ctrl/cmd+arrowDown command from the neume.

**Consideration:** 

Without this fix, it is possible to bridge over breath marks and crosses with western-style melisma using this entry method.  

With this fix applied, it will be necessary for the user to change the "Accepts Lyrics" setting for the neume to "Melisma Only" or "Yes" in order to bridge the symbol with a melisma, which matches the behavior we see when using the lyrics toolbar.

I had considered applying this behavior to the "Melisma Only" characters as well, but I thought this might cause problems in some cases, and I wasn't sure how to get it to render the melisma as it was passing over the note.

### DIFF
--- a/src/components/Editor.vue
+++ b/src/components/Editor.vue
@@ -3994,7 +3994,14 @@ export default class Editor extends Vue {
       // Find the index of the next note
       for (let i = currentIndex + 1; i < this.elements.length; i++) {
         if (this.elements[i].elementType === ElementType.Note) {
-          return i;
+          if (
+            this.lyricService.getEffectiveAcceptsLyrics(
+              this.elements[i] as NoteElement,
+              null,
+            ) != AcceptsLyricsOption.No
+          ) {
+            return i;
+          }
         }
       }
     }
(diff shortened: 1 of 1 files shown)


danielgarthur:
I'm still thinking about this, but I think if anything is to be done (and this seems like a very minor bug), it is that the breath/cross need to be somehow further demoted. I'm not sure what that looks like because I would still want to preserve the ability to put lyrics under them. For example, maybe someone is making educational material and wants to label the neume.

PR #1377, opened 2025-04-18
Add macOS ARM64 build

### LINKED ISSUE #1371: Playback hiccups
**This bug was found in**

- [x] Desktop
- [ ] Web App
- [ ] Both

**Describe the bug**
I'm not sure this is reproducible on other computers or operating systems, but since recently playback with every single Neanes file I tried has experienced constant hiccups throughout the entire piece. I do have enough disk space, so this shouldn't be an issue, and I believe neither memory (8 Gb) is an issue. Anybody else with similar experiences?

- OS: macOS Sequoia 15.4
- Version: 0.4.81

basil:
Reverts neanes/neanes@cea1b32745e38b13496ab45f5b39c05c256d2a2f. Fixes #1371.

### Testing done

With `CSC_FOR_PULL_REQUEST` temporarily set to `true`, observed the expected

```
  • falling back to ad-hoc signature for macOS application code signing
  • signing         file=dist/mac-arm64/Neanes.app platform=darwin type=distribution identityName=- identityHash=none provisioningProfile=none
  • skipped macOS notarization  reason=`notarize` options were unable to be generated
```

from https://github.com/electron-userland/electron-builder/pull/9007. Downloaded the resulting `.dmg` file and installed to `/Applications`. Right-clicked and selected **Open,** saw "Neanes Not Opened: Apple could not verify Neanes is free of malware that may harm your Mac or compromise your privacy." (This is expected, and is better than the previous result of "damaged.") Went to **Settings, Privacy & Security,** saw "Neanes was blocked to protect your Mac," and clicked **Open Anyway.** Neanes launched and playback was fine.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-13]
+        os: [ubuntu-latest, windows-latest, macos-latest, macos-13]
     steps:
       - uses: actions/checkout@v4
 
@@ -41,3 +41,10 @@ jobs:
       - run: npm run build
         env:
           GH_TOKEN: ${{secrets.GITHUB_TOKEN}}
+
+      - name: Publish
+        uses: actions/upload-artifact@v4
+        with:
+          name: Neanes-${{ matrix.os }}
+          path: dist/Neanes-*
+          retention-days: 3
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "@vue/eslint-config-prettier": "^10.2.0",
     "@vue/eslint-config-typescript": "^14.5.0",
     "electron": "^35.1.5",
-    "electron-builder": "^26.0.12",
+    "electron-builder": "^26.0.13",
     "electron-updater": "^6.6.2",
     "eslint": "^9.24.0",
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -41,3 +41,10 @@ jobs:
       - run: npm run build
         env:
           GH_TOKEN: ${{secrets.GITHUB_TOKEN}}
+
+      - name: Publish
+        uses: actions/upload-artifact@v4
+        with:
+          name: Neanes-${{ matrix.os }}
+          path: dist/Neanes-*
+          retention-days: 3
```
basil, line 50: Not strictly needed for this PR, but needed for me to be able to test the CI build and seems useful more generally. Signing will not take place without setting `CSC_FOR_PULL_REQUEST=true`, and I used that during testing but did not retain it in the final version of this PR because of scary-looking security warnings printed to the logs. `CSC_FOR_PULL_REQUEST` should not be necessary for the release pipeline.

package.json
```diff
@@ -65,7 +65,7 @@
     "@vue/eslint-config-prettier": "^10.2.0",
     "@vue/eslint-config-typescript": "^14.5.0",
     "electron": "^35.1.5",
-    "electron-builder": "^26.0.12",
+    "electron-builder": "^26.0.13",
```
basil, line 68: https://github.com/electron-userland/electron-builder/pull/9007

danielgarthur (approved):

PR #1416, opened 2025-05-01
Update transitive dependencies

basil:
Checking if this resolves any of the alerts in https://github.com/neanes/neanes/security/dependabot.


basil:
Looks like it did resolve five of them.

PR #1456, opened 2025-05-22
Fix document title bug when changing tabs

danielgarthur:
This fixes a bug where the window title doesn't change when switching tabs unless the workspace has been saved at least once.


danielgarthur:
This contained a bug. Fixed in https://github.com/neanes/neanes/commit/3caa69f9b6aa8fd80c966417a5e439eac0212c61

PR #1522, opened 2025-06-24
Add mode key for Plagal of First Mode as from Pa

### LINKED ISSUE #1521: Mode Key Signature: Pl 1 from KE with PA fthora
When the option off adding fthoras to the innitial Martyria is activated, not all of the martyria gets one, for instance pl. 1 from Ke does not get its fthora PA.

Also, Could you add a possibility to create text boxes next to signs ? Sometimes there is enough space to write a whole verse on the same line as the musical notation, because it brakes the line and there are only 2 or 3 signs.

Thank you for this fantastic software.

Richard

basil:
Fixes #1521

image


basil:
I think I prefer the first example from https://github.com/neanes/neanes/issues/1521#issuecomment-3001193397.

danielgarthur:
> @danielgarthur Any suggestions on how to represent this in Neanes? I don't think the existing `ModeKey.vue` supports this use case. Should I add a new `quantitativeNeumeLeft`/`fthoraAboveQuantitativeNeumeLeft` to go along with `quantitativeNeumeRight`/`fthoraAboveQuantitativeNeumeRight`?

Sounds good to me.

danielgarthur (approved):

Issue #430, opened 2023-11-28
Pitch upheavals in playback


basil:
The function in the playback service to compute the transposition moria works most of the time but not all of the time. It currently works by computing the distance from the physical note to Di in the old scale, switching scales, then computing the distance from Di to the virtual note in the new scale, then summing the two values. This is a heuristic that I settled on after a lot of empirical experimentation. This heuristic works in many cases, but it does not work in all cases. Good stress tests are the Anoixantaria by Koukouzelis (abbreviated by Chourmouzios) and "To tritto tis erotiseos", both of which currently result in incorrect transposition moria being calculated.

[This very old translation of the Melodos manual](http://graeca.mrezha.net/upload/MontrealPsaltiki/Melodos/Melodos_2007_v_003_gr_en.htm) shows one approach to the problem in the appendix, which is worth quoting in its entirety:

<details><summary><strong>Appendix A: Fthores</strong></summary>

> For a better understanding of how to use fthores, the following example is provided (the melody from which these examples is drawn is provided with Melodos). The hymn used is "Awed by the beauty," which is chanted at the end of the Akathist service.
> 
> The examined fragment is the phrase «διό ως προσετάγην». Here is the classical line:
> 
> ![image](https://github.com/danielgarthur/neanes/assets/29850/1bf93d5f-aa9b-4aba-abd3-94edcebbd3e1)
> 
> In the beginning of this example, the melody has already been modulated and is in hard chromatic. On the syllable «διό», a diatonic Ga fthora has been added to the note Ga. If this modulation is done without any other predisposition, then the scales appearing on the right to the score will present as follows:
> 
> ![image](https://github.com/danielgarthur/neanes/assets/29850/3eebd500-9cd1-45a4-89a6-44fe3cd1449e)
> 
> The three scales presented correspond to the following: (1) initial orange scale (initial hard chromatic scale), (2) middle blue scale (entire modulated diatonic scale) and (3) final blue scale (partially [chroa-] or entirely [fthora-] modulated scale (this is the scale that will be perforemed by Melodos’ playback module). The modulation window shown above shows that a chromatic Ga has been modulated to a diatonic Ga. However, the user will note that the limits of the chromatic and diatonic pentachords do not coincide, even though the total number of units is the same for both these pentachords (42): Pa and Ke of the hard chromatic scale are not of the same pitch in the diatonic scale.
> 
[...]

PR #1534, opened 2025-06-28
Fix specific case of spathi playback

danielgarthur:
This PR updates the playback service to remember the last pivot used. If the current or previous scale is Spathi on KE, for example, the pivot is KE.

Without this PR, KE is only used as the pivot when switching **to** spathi, but not when switching **away** from spathi.

This fixes the spathi example in #430.


danielgarthur:
Closed in favor of #1539

PR #1539, opened 2025-07-01
Messing around with upheavals

danielgarthur:


basil (approved):
Very nice! Impressive how well all these edge cases are handled. I'm not sure what Melodos playback is like these days, but this very well could be the state of the art for Byzantine music playback software.

While testing my _Axion Estin_ scores, I noticed that in line 9 of the Plagal Second Mode setting by Triantaphyllus Georgiades the ison drops a little too low. But I compared that to the playback before this PR, and this PR is not making things any worse than before. So that isn't a regression, and I can file a separate ticket about it. In any case, it's an extremely minor issue that I wouldn't worry too much about. I am just mentioned it here for completeness, since we are discussing a variety of edge cases as part of this issue/PR.

PR #1570, opened 2025-07-08
Show only font family in font family selection dropdown

basil:
List font families using the [local font access API](https://developer.mozilla.org/en-US/docs/Web/API/Local_Font_Access_API), which is a reliable way to get list of font families in Chromium.


basil:
As an aside, this seems to allow custom fonts to be used in the web application in Brave after accepting the prompt asking for permission to do so. In Firefox, it didn't work as expected. The web application in Chromium is a bit of a strange case: I got `SecurityError: User activation is required` and was never even prompted for permission. :man_shrugging:

basil:
> I like this so far, but there needs to be a way to select styles, both in CKEditor and elsewhere. And some amount of backwards compatibility, ideally, so that if someone has a score with `Extra Light` in the font name, it is still the select font and style when opening in the new version of the app.

⇨ #2296

PR #1665, opened 2025-08-19
Prevent invalid neume combinations

ilizol:


basil:
What problem is this trying to solve, and how was it tested?

ilizol:
> What problem is this trying to solve, and how was it tested?

Screenshot_2025-08-19-07-05-30-030_com android chrome_20250820034720

Before the change it was allowed the antikenoma to be paired with oligon and kentimata above combinations, that orthographically is not correct.
After the change the antikenoma is not inserted after clicking the antikenoma button for these combinations.

ilizol:
Screenshot_2025-08-20-04-22-03-475_com android chrome_20250820042414

The apli should not be allowed to be paired with the petasti plus running elafron combination if antikenima is not allowed.
Also the apli anchor should be removed from the petasti plus running elafron combination from the sbmufl fonts.

PR #1670, opened 2025-08-20
Prevent invalid neume combinations (apli)

ilizol:
`src/models/NeumeReplacements.ts`
`timeReplacementMap`
`TimeNeume.Hapli`

https://github.com/neanes/neanes/pull/1665#issuecomment-3203793601


ilizol:
@danielgarthur 
@basil 

Please review.

danielgarthur:
Is this really correct? Is the issue rather that the running elaphron should allow the antikenoma? It's probably a rare combination but still valid since all characters with a petaste can take apli with antikenoma, I think.

PR #1757, opened 2025-10-14
Use Options API

danielgarthur:


basil:
Caused #2241, #2242, and #2243

PR #2016, opened 2026-03-25
Expand touchscreen support

danielgarthur:
This PR introduces a configurable option in the editor preferences dialog to control toolbar menu interaction behavior. Users can now switch between the existing interaction model and a new click-to-open mode, where menus open on click and stay open until an item is chosen with a second click, or the user clicks away.

### DIFF
--- a/src/components/ButtonWithMenu.vue
+++ b/src/components/ButtonWithMenu.vue
@@ -5,10 +5,15 @@
     @mouseleave="handleMouseLeave"
   >
     <button class="neume-button" :disabled="disabled">
-      <img draggable="false" :src="mainIcon" v-if="mainIcon" />
+      <img
+        draggable="false"
+        :src="mainIcon"
+        v-if="mainIcon"
+        :style="imgStyle"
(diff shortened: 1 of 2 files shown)


basil (commented):
Is there a particular reason the menu interaction mode needs to be a user preference? The two modes don't seem to conflict: hold/drag/release works by tracking `mouseenter` during a hold and selecting on `mouseup`, while click to open works by keeping the menu open when `mouseup` has no selected item and handling the subsequent click. Since the distinguishing factor is simply whether the user releases over a menu item or over the button itself, I wonder if both behaviors could be enabled simultaneously.

basil:
Perhaps we could track whether the mouse has actually moved since the menu opened with a flag? That way, the item that appears under the cursor when the menu opens wouldn't be treated as selected, since the cursor didn't move to it. If it helps, I already spent the tokens on this here (but I don't have a touchscreen to test in that environment): https://github.com/basil/neanes/tree/touchscreen

PR #2020, opened 2026-03-26
Stop melisma underscore from extending past measure bars

basil:
The melisma end position was calculated using `neumeWidth`, which includes the measure bar right width. This caused the melisma underscore to visually extend past barlines. Now we subtract the measure bar width from the end position for all melisma types (hyphen, underscore, and RTL).

image

Stacked PR; without #2019, the old code happened to inflate `neumeWidth` with the computed measure bar.


basil:
Thanks! I cut a release just now, and while I was at it, I cleaned up some of the CI/release jobs as well.

PR #2061, opened 2026-04-02
feat: implement Knuth-Plass optimal line breaking

### LINKED ISSUE #90: [ENHANCEMENT] Knuth & Plass line breaking algorithm
See [_Breaking Paragraphs into Lines_](http://www.eprg.org/G53DOC/pdfs/knuth-plass-breaking.pdf) by Donald Knuth & Michael Plass (1981). Serious typesetting programs that deal with justification, including TeX and Adobe InDesign, use a sophisticated algorithm for breaking paragraphs into lines. In the abovementioned paper, Knuth & Plass present a model of boxes, glue, and penalty specifications to efficiently compute the optimal set of line breaks. The Adobe Paragraph Composer works in much the same way. This avoids excessive hyphenation and rivers of white space in the justified output.

I have never seen anyone attempt to apply this concept to Western notation. About ten years ago I was using Finale, and I had to manually break the score into lines to get good output. I have never used Dorico and have no idea if it is any better.

Applying Knuth & Plass line breaking to Western notation music seems fairly difficult due to the geometric placement of notes on the Western staff, but it seems plausible that this algorithm could be applied to Byzantine notation music. The goal would be to avoid the bad line and page breaks that plague poorly typeset music. For example, most sticheraric and slow heirmologic pieces have a duple rhythm of 4-beat or 6-beat measures. Naïvely breaking the line between an odd-numbered beat and an even-numbered beat is undesirable, so this could be given a worse penalty specification in Knuth & Plass's algorithm. Similarly, a gorgon applies to the neume it is placed on as well as the preceding neume, so a line or page break between these two neumes is also undesirable.

basil:
Replace the greedy first-fit line breaking algorithm with the Knuth-Plass optimum-fit algorithm, producing beautifully even spacing across every line, like TeX and Adobe InDesign.

The LayoutService now encodes each paragraph as a sequence of boxes, glue, and penalties, then uses dynamic programming to find the optimal set of breakpoints that minimizes the squared deviation from ideal spacing. Lyrics are handled with the Hegazy-Gourlay approach (as in LilyPond), and TeX's Restriction 1 is relaxed to support negative widths arising from neume/lyric geometry.

Fixes #90

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ Neanes (pronounced neh-ah-ness) is a free and open source scorewriter for notati
 - Rapid entry of neumes and lyrics
 - Automatic alignment of supporting neumes (i.e. fthoras, accidentals, klasmas, gorgons, et al)
 - Automatic calculation of martyria
+- [Optimal line breaking](notes/knuth_plass.md) for beautifully even spacing across every line, just like TeX and Adobe InDesign
 - Print or export to PDF
 - Export to HTML using [ByzHtml](https://danielgarthur.github.io/byzhtml) web components
 - Export to PNG
--- a/docs/index.md
+++ b/docs/index.md
@@ -27,6 +27,8 @@ features:
     details: Point and click to add neumes and see how they will appear on the page.
   - title: Rapid Entry of Neumes and Lyrics
     details: Neumes and lyrics are automatically aligned and martyria are automatically calculated.
+  - title: Advanced Typography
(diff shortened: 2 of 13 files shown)


basil:
Sorry, I accidentally pushed this to `master`, which closed this PR. I fixed that by force-pushing backward (restored `master`) and will open a new PR to continue this one.

basil:
Continued in #2070

PR #2070, opened 2026-04-08
Knuth-Plass line breaking algorithm

### LINKED ISSUE #90: [ENHANCEMENT] Knuth & Plass line breaking algorithm

basil:
Replace the greedy first-fit line breaking algorithm with the Knuth-Plass optimum-fit algorithm, producing beautifully even spacing across every line, like TeX and Adobe InDesign.

The LayoutService now encodes each paragraph as a sequence of boxes, glue, and penalties, then uses dynamic programming to find the optimal set of breakpoints that minimizes the squared deviation from ideal spacing. Lyrics are handled with the Hegazy-Gourlay approach (as in LilyPond), and TeX's Restriction 1 is relaxed to support negative widths arising from neume/lyric geometry.

Fixes #90


basil:
I think we should consider removing the metrics Python script and metrics collection part of this PR. It was hastily vibe-coded and I haven't really thought critically if the numbers are fully correct. And since it's off by default, it seems likely to be a maintenance burden or source of regressions. I think we could always re-add it as part of a future PR if we miss it. I don't feel strongly either way, though: happy to keep it if you think it is useful.

danielgarthur:
I'm fine with removing it.

basil:
FYI I wrote a blog post about the supporting work done in the Knuth-Plass linebreak library to support this project: https://basilcrow.com/blog/relaxing-knuths-restrictions/

notes/knuth_plass.md
```diff
@@ -0,0 +1,541 @@
...
+### Lyricless scores
+
+Consider first a lyricless score of Byzantine music.
+Each neume group, for example a simple oligon or an ison with kentimata over a supporting oligon, and each martyria can be modeled as a box whose width is simply the width of the notated group itself.
+The space between ordinary neumes is modeled as glue with a user-configurable width that may stretch or shrink by up to one half of its preferred width.
+
+Following several classical 19th-century publications, ordinary martyriae are given extra elasticity.
+When a martyria sits between neighboring elements, the surrounding martyria glues together contribute an additional 2/5 em of stretch, that is, 1/5 em on each side, where 1 em is the neume font size.
+This lets the martyria absorb extra line slack before the surrounding neumes do, which helps keep ordinary neume spacing more even from line to line.
+After the martyria box, the code inserts a zero-width pre-break glue, then a zero-cost break penalty, and finally a single martyria glue whose preferred width is the ordinary martyria spacing plus the fixed trailing padding.
```
danielgarthur, line 141: What is the purpose of the zero width glue and zero width penalty inserted before the post break glue? I have been pondering for a while but I can't see any functional difference if they were omitted.

danielgarthur: I do recognize this `penalty(0, MAX_COST, 0) glue(0, 0, 0) penalty(0,0,0)` as the normal-form from the Knuth-Plass paper. It's not strictly necessary, though, right?

basil:
My changes from https://basilcrow.com/blog/relaxing-knuths-restrictions/ were merged and released in the upstream `tex-linebreak` library, so I've dropped the fork from this changeset.

danielgarthur (commented):
Apologies. I never "completed the review". 

I just had a couple of questions to make sure that I understand everything. I'm not trying to nitpick the implementation. I want to make sure that I'm not missing important details.

notes/knuth_plass.md
basil: @danielgarthur You are right: for the martyria case, the zero-width pre-break glue and zero-width break penalty are not intrinsically needed. They are just to preserve the conventional normal form from the Knuth-Plass paper.

We can't simply delete only the `glue(0)` and `penalty(0)` from the current sequence, though, because the preceding `penalty(MAX_COST)` would leave the post-break glue preceded by a penalty, not a box, so it would no longer be a legal glue breakpoint. But if we replaced the whole encoding with just `<martyriaBox> <postBreakGlue>` then the glue itself would be a legal breakpoint.

While the above would be functionally equivalent in line metrics, I still feel that the explicit normal form is useful because it keeps martyriae using the same consistent breakpoint encoding as notes and leaves room for future nonzero break cost/width without changing the design.

danielgarthur: Okay, I understand now. Thanks.

danielgarthur:
@basil If nothing has drastically changed since April 8th or so, then I'm good to merge this in.

basil:
Thanks @danielgarthur. Apart from rebasing against latest `HEAD`, nothing has changed in the list of commits. The only recent addition is commit 77225e56a7c5f12fb979f63bd599b51fb7693134 to fix a bug I noticed while working on a recent project. Apart from that things have been stable and I have been using this branch as my daily driver in addition to converting my web site over to it.

Since this is a large change, I'll wait a few more days before merging this, or for you to merge it, whichever comes first. Feel free to let me know if you have any other questions or feedback during that time. If you do merge it, I would prefer that it would be squashed into one line without all the intermediary commit messages, since those often reflect an internal state of mind.

danielgarthur:
I'll let you go ahead and merge it when you are ready.

PR #2142, opened 2026-05-18
fix: Move electron-builder output out of the asar input directory

### LINKED ISSUE #2141: Neanes broken in 0.5.9 for macOS arm64
Neanes won't even open starting with this release. It seems to try to start, then immediately abandons. I don't even see the window open. 

I've gone back to 0.5.8 and this is working fine.

**Desktop Version:**

- OS: macOS Tahoe 26.3.1
- Neanes Version 0.5.9 and 0.5.10

basil:
Fixes #2141

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ yarn-error.log*
 #Electron-builder output
 /dist
 /dist-electron
+/release
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,6 +1,7 @@
 .vscode
 dist
 dist-electron
+release
 docs/.vitepress/cache
 node_modules
 package-lock.json
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,7 +23,7 @@ To create a release build, type:
 npm run build
 ```
 
-The build artifacts are found in `dist/`. For Windows, the installer will be called `Neanes Setup [version].exe`. The raw files can be found in `dist/win_unpacked`.
+The build artifacts are found in `release/`. For Windows, the installer will be called `Neanes Setup [version].exe`. The raw files can be found in `release/win-unpacked`.
 
 ### Development
 
--- a/electron-builder.json
+++ b/electron-builder.json
@@ -1,6 +1,9 @@
 {
   "appId": "com.danielgarthur.neanes",
   "files": ["dist-electron", "dist"],
+  "directories": {
+    "output": "release"
+  },
   "productName": "Neanes",
   "copyright": "Copyright © 2021 danielgarthur",
   "publish": ["github"],
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -5,7 +5,12 @@ import pluginVue from 'eslint-plugin-vue';
 
 export default [
   {
-    ignores: ['dist/**/*', 'dist-electron/**/*', '**/.vitepress/cache/**/*'],
+    ignores: [
+      'dist/**/*',
+      'dist-electron/**/*',
+      'release/**/*',
+      '**/.vitepress/cache/**/*',
+    ],
   },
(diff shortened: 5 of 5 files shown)


danielgarthur:
@basil I'm confused. What caused this to break and why do we now need to do all this?

danielgarthur:
@basil When I run `npm run build` the resulting arm64 DMG works, but when I add `npm run build -- --x64 --arm64` it fails. So I think it's related to that change. I'm not sure what the fix is, yet.

basil:
@danielgarthur The macOS arm64 releases for Neanes 0.5.9 and 0.5.10 failed at startup because the packaged `app.asar` was corrupt. The asar header still recorded the expected metadata for the root `package.json`, but the bytes at that recorded offset were not `package.json`; they were content from unrelated packaged files. Electron 42 verifies asar file integrity while loading the app, so the process exited before any window was created. The Intel macOS build, Linux build, and Windows build were not affected in the same way.

`electron-builder.json` included this configuration:

```json
"files": ["dist-electron", "dist"]
```

That is partly correct: Vite writes the renderer bundle to `dist/`, and that bundle needs to be included in the app. The problem is that electron-builder also used `dist/` as its default output directory. That meant the directory being packaged into the asar was also the directory where electron-builder was writing installers, zips, unpacked apps, blockmaps, and other build output.

This had been a latent configuration bug for a while. It became release-breaking when the macOS publish job was changed to build both architectures in one electron-builder invocation:

```sh
npm run release -- --x64 --arm64
```

That single invocation is the right logic for producing one `latest-mac.yml` containing both Intel and Apple Silicon update entries. However, with electron-builder output still going into `dist/`, the second architecture could see output produced by the first architecture and include that output in its own asar. In 0.5.9 and 0.5.10, the arm64 asar picked up macOS build artifacts from the shared `dist/` tree.

Electron 42 then made the corruption fatal. It verifies per-file asar integrity when loading files from the archive. Because the data at the `package.json` offset did not match the hash recorded in the asar header, Electron aborted during startup before the app could show a window.

In the combined macOS build, x64 was packaged first and arm64 second. The x64 package was produced before there was much electron-builder output to accidentally include. By the time arm64 packaging ran, the shared output directory already contained the x64 zip, DMG, and unpacked app. The arm64 asar therefore absorbed output from the earlier architecture.

The fix is to move electron-builder's output directory out of the directory that is packaged into the asar:

```json
"directories": {
  "output": "release"
}
```

With this change:

- `dist/` remains Vite's renderer output and is still included in the app.
- `dist-electron/` remains the Electron main/preload output and is still included in the app.
- `release/` becomes electron-builder's output directory and is not included in the asar.

The combined `--x64 --arm64` macOS build form is retained. It was not the root bug; it only exposed the existing directory collision. Keeping the combined build is still important so macOS update metadata can describe both architectures in one `latest-mac.yml`.

## Testing done

The PR CI run on macOS built both architectures successfully in one command:

```sh
npm run build -- --x64 --arm64
```

The macOS log showed electron-builder writing packaged apps to the new output
directory:

- `appOutDir=release/mac`
- `appOutDir=release/mac-arm64`

It also showed all expected macOS artifacts being produced under `release/`:

- `release/Neanes-0.5.10-mac.zip`
- `release/Neanes-0.5.10.dmg`
- `release/Neanes-0.5.10-arm64-mac.zip`
- `release/Neanes-0.5.10-arm64.dmg`

During validation, an asar check was run against the macOS CI output. It confirmed that both packaged asars had a valid root `package.json` at the recorded asar offset and did not contain electron-builder output paths:

- `release/mac/Neanes.app/Contents/Resources/app.asar`
- `release/mac-arm64/Neanes.app/Contents/Resources/app.asar`

danielgarthur:
@basil Thanks for the explanation and for the fix.

PR #2149, opened 2026-05-18
fix: Page break adds an extra page

### LINKED ISSUE #2159: page break not working now
I see that Neanes just updated.  After doing so, I started typing a new score.  When I clicked on the page break to put that part on the next page, it didn't work.  It indented that part in more to the left, with nothing after it on the next page.  Please look at this issue and I hope it can be corrected.  Thank you...

Christopher Holwey

danielgarthur:
## Problem and Reproduction Steps

- Create a new score. 
- Delete the text box and mode key for simplicity. 
- Add a single oligon. 
- Put a page break on the oligon. 
- No page is added.

There is an error in the console related to out of bounds access to an array. If this is fixed, then instead of two pages, three pages are shown. The first page contains the oligon. The second nothing. The third page contains the empty element.

### DIFF
--- a/src/services/LayoutService.ts
+++ b/src/services/LayoutService.ts
@@ -1196,6 +1196,10 @@ export class LayoutService {
           pages.push(page);
           currentPageHeightPx = lastLineHeightPx;
 
+          if (lastElementWasPageBreak) {
+            lastElementWasPageBreak = false;
+          }
+
           // Recalculate the height of the headers/footers of the new page
           ({ extraHeaderHeightPx, extraFooterHeightPx } =
             this.getExtraHeaderFooterHeight(score, pageSetup, pages.length));
@@ -3052,9 +3056,16 @@ export class LayoutService {
 
         let firstElementOnNextLine: ScoreElement | null = null;
 
-        if (lineIndex + 1 < page.lines.length) {
+        if (
+          lineIndex + 1 < page.lines.length &&
+          page.lines[lineIndex + 1].elements.length > 0
+        ) {
           firstElementOnNextLine = page.lines[lineIndex + 1].elements[0];
-        } else if (pageIndex + 1 < pages.length) {
+        } else if (
+          pageIndex + 1 < pages.length &&
+          pages[pageIndex + 1].lines.length > 0 &&
+          pages[pageIndex + 1].lines[0].elements.length > 0
+        ) {
           firstElementOnNextLine = pages[pageIndex + 1].lines[0].elements[0];
         }
 
--- a/src/services/LayoutService.ts
+++ b/src/services/LayoutService.ts
@@ -1196,9 +1196,7 @@ export class LayoutService {
           pages.push(page);
           currentPageHeightPx = lastLineHeightPx;
 
-          if (lastElementWasPageBreak) {
-            lastElementWasPageBreak = false;
-          }
+          lastElementWasPageBreak = false;
 
           // Recalculate the height of the headers/footers of the new page
           ({ extraHeaderHeightPx, extraFooterHeightPx } =
--- a/src/services/LayoutService.ts
+++ b/src/services/LayoutService.ts
@@ -1196,6 +1196,11 @@ export class LayoutService {
           pages.push(page);
           currentPageHeightPx = lastLineHeightPx;
 
+          // Consume the page-break trigger. Subsequent positioned items
+          // belonging to the same break (glues, penalties without an
+          // associated element) hit the `continue` below and skip the
+          // `lastElementWasPageBreak` update, so without this reset the
+          // flag re-fires this branch and leaves an empty page behind.
           lastElementWasPageBreak = false;
 
           // Recalculate the height of the headers/footers of the new page
@@ -1547,7 +1552,15 @@ export class LayoutService {
         previousLyricsEndPx -
         workspace.neumesEndPx +
         pageSetup.neumeDefaultSpacing;
-      this.addAnonymousBox(adjustment, workspace);
+
+      // A zero-width spacer would only add a box to pendingParagraph without
+      // advancing neumesEndPx. The immediately-following addBox call already
+      // supplies the box that anchors any later glue as a legal breakpoint,
+      // so the spacer is redundant when adjustment is 0. This mirrors the
+      // martyria bar-transfer call site, which is also guarded by > 0.
+      if (adjustment > 0) {
+        this.addAnonymousBox(adjustment, workspace);
+      }
     }
 
     workspace.lyricsEndPx =


src/services/LayoutService.ts
```diff
@@ -1196,6 +1196,8 @@ export class LayoutService {
           pages.push(page);
           currentPageHeightPx = lastLineHeightPx;
 
+          lastElementWasPageBreak = false;
```
danielgarthur, line 1204: @basil I'm not sure whether this is the best fix or if it has maybe something to do with how the EmptyElement is handled.

basil:
Not sure what this is about. I tried to reproduce the issue but I only see two pages.

danielgarthur:
> Not sure what this is about. I tried to reproduce the issue but I only see two pages.

So on the latest `master` branch, if you open a new score, add an ison and a page break, you don't see an error in the console?

image

basil (commented):
@danielgarthur I see, I can reproduce that. I think the fix should be much simpler:

```ts
diff --git a/src/services/LayoutService.ts b/src/services/LayoutService.ts
index eb9b5673..508310d6 100644
--- a/src/services/LayoutService.ts
+++ b/src/services/LayoutService.ts
@@ -1196,6 +1196,13 @@ export class LayoutService {
           pages.push(page);
           currentPageHeightPx = lastLineHeightPx;

+          // Consume the page-break trigger. Subsequent positioned items
+          // belonging to the same break (glues, penalties without an
+          // associated element) hit the `continue` below and skip the
+          // `lastElementWasPageBreak` update, so without this reset the
+          // flag re-fires this branch and leaves an empty page behind.
+          lastElementWasPageBreak = false;
+
           // Recalculate the height of the headers/footers of the new page
           ({ extraHeaderHeightPx, extraFooterHeightPx } =
             this.getExtraHeaderFooterHeight(score, pageSetup, pages.length));
```

The throwing expression is:

```ts
firstElementOnNextLine = pages[pageIndex + 1].lines[0].elements[0];
```

`pages[pageIndex + 1].lines` is the empty array `[]`, so `.lines[0]` is `undefined`, and reading `.elements` on it throws.

When I took a look at this, I saw that two firings of the `requiresNewPage` branch happened back-to-back for a single user page break, with `currentPageLineCount` unchanged (no new line added between them). The second firing popped the only line off the freshly-created page, leaving page 1 empty. `addMelismas` then crashed when it reached forward from page 0 into the empty page 1.

`processPages` walks a list of positioned items produced by the Knuth-Plass pass. Each item may be a box (an element box, the only kind that carries an `element` field), a glue, or a penalty. The new-page check runs every iteration:

```ts
const requiresNewPage =
  currentPageHeightPx + additionalHeight > innerPageHeight ||
  lastElementWasPageBreak;
```

`lastElementWasPageBreak` is updated near the bottom of the loop body:

```ts
lastElementWasPageBreak = element.pageBreak;
```

But earlier in the body, items without an associated element short-circuit out:

```ts
if (!('element' in item)) {
  continue;
}
const element = (item as ElementBox).element;
```

Glue and penalty items belonging to the same break therefore skip the update and leave `lastElementWasPageBreak` set to `true` from the prior page-break element. On the next iteration, `requiresNewPage` fires again from the stale flag, pops the only line off the page that was just created, and produces an empty page.

The Knuth-Plass refactor introduced these non-element items into the loop; the pre-refactor loop did not have positioned items without elements, so the flag's lack of explicit consumption was never observable. The fix is to consume the page-break trigger inside the new-page branch.

danielgarthur:
> @danielgarthur I see, I can reproduce that. I think the fix should be much simpler:
> 
> ```ts
> diff --git a/src/services/LayoutService.ts b/src/services/LayoutService.ts
> index eb9b5673..508310d6 100644
> --- a/src/services/LayoutService.ts
> +++ b/src/services/LayoutService.ts
> @@ -1196,6 +1196,13 @@ export class LayoutService {
>            pages.push(page);
>            currentPageHeightPx = lastLineHeightPx;
> 
> +          // Consume the page-break trigger. Subsequent positioned items
> +          // belonging to the same break (glues, penalties without an
> +          // associated element) hit the `continue` below and skip the
> +          // `lastElementWasPageBreak` update, so without this reset the
> +          // flag re-fires this branch and leaves an empty page behind.
> +          lastElementWasPageBreak = false;
> +
>            // Recalculate the height of the headers/footers of the new page
>            ({ extraHeaderHeightPx, extraFooterHeightPx } =
>              this.getExtraHeaderFooterHeight(score, pageSetup, pages.length));
> ```
> 
> The throwing expression is:
> 
> ```ts
> firstElementOnNextLine = pages[pageIndex + 1].lines[0].elements[0];
> ```
> 
> `pages[pageIndex + 1].lines` is the empty array `[]`, so `.lines[0]` is `undefined`, and reading `.elements` on it throws.
> 
> When I took a look at this, I saw that two firings of the `requiresNewPage` branch happened back-to-back for a single user page break, with `currentPageLineCount` unchanged (no new line added between them). The second firing popped the only line off the freshly-created page, leaving page 1 empty. `addMelismas` then crashed when it reached forward from page 0 into the empty page 1.
> 
> `processPages` walks a list of positioned items produced by the Knuth-Plass pass. Each item may be a box (an element box, the only kind that carries an `element` field), a glue, or a penalty. The new-page check runs every iteration:
> 
> ```ts
> const requiresNewPage =
>   currentPageHeightPx + additionalHeight > innerPageHeight ||
>   lastElementWasPageBreak;
> ```
> 
> `lastElementWasPageBreak` is updated near the bottom of the loop body:
> 
> ```ts
> lastElementWasPageBreak = element.pageBreak;
> ```
> 
> But earlier in the body, items without an associated element short-circuit out:
> 
> ```ts
> if (!('element' in item)) {
>   continue;
> }
> const element = (item as ElementBox).element;
> ```
> 
> Glue and penalty items belonging to the same break therefore skip the update and leave `lastElementWasPageBreak` set to `true` from the prior page-break element. On the next iteration, `requiresNewPage` fires again from the stale flag, pops the only line off the page that was just created, and produces an empty page.
> 
> The Knuth-Plass refactor introduced these non-element items into the loop; the pre-refactor loop did not have positioned items without elements, so the flag's lack of explicit consumption was never observable. The fix is to consume the page-break trigger inside the new-page branch.

Thanks. I left the bounds checks in there just because it felt safer, but they are not strictly necessary and can be removed again for simplicity.

As for the anonymous box, can it be omitted if the width is 0?

basil:
I think so. I tested this:

```ts
diff --git a/src/services/LayoutService.ts b/src/services/LayoutService.ts
index eb9b5673..7d4a3cd3 100644
--- a/src/services/LayoutService.ts
+++ b/src/services/LayoutService.ts
@@ -1545,7 +1545,14 @@ export class LayoutService {
         previousLyricsEndPx -
         workspace.neumesEndPx +
         pageSetup.neumeDefaultSpacing;
-      this.addAnonymousBox(adjustment, workspace);
+      // A zero-width spacer would only add a box to pendingParagraph without
+      // advancing neumesEndPx. The immediately-following addBox call already
+      // supplies the box that anchors any later glue as a legal breakpoint,
+      // so the spacer is redundant when adjustment is 0. This mirrors the
+      // martyria bar-transfer call site, which is also guarded by > 0.
+      if (adjustment > 0) {
+        this.addAnonymousBox(adjustment, workspace);
+      }
     }

     workspace.lyricsEndPx =
```

and I didn't encounter any problems with my scores.

PR #2154, opened 2026-05-18
feat: Ignore punctuation when positioning lyrics

### LINKED ISSUE #2031: Ignore punctuation at start and end of lyric when positioning
By default, punctuation at the start and end of a lyric should be ignored when positioning. From Dorico's default configuration:

Image

danielgarthur:
## Before
image

## After

image

### DIFF
--- a/src/components/PageSetupDialog.vue
+++ b/src/components/PageSetupDialog.vue
@@ -797,23 +797,6 @@
                 $t('dialog:pageSetup.disableGreekMelismataDescription')
               }}</label>
             </div>
-            <div class="form-group">
-              <label class="name">{{
-                $t('dialog:pageSetup.lyricsMelismaCutoffWidth')
-              }}</label>
-              <div class="description">
-                {{ $t('dialog:pageSetup.lyricsMelismaCutoffWidthDescription') }}
-              </div>
-              <InputUnit
-                v-model="form.lyricsMelismaCutoffWidth"
-                class="unit-input"
-                unit="pt"
-                :min="0"
-                :step="1"
-                :precision="0"
-              />
-            </div>
-
             <div ref="dropCapsRef" class="subheader">
               {{ $t('dialog:pageSetup.dropCaps') }}
             </div>
@@ -996,6 +979,42 @@
                   class="drop-caps-input"
                 />
               </div>
+            </div>
+            <div class="form-group">
+              <label class="name">{{
+                $t('dialog:pageSetup.lyricsMelismaCutoffWidth')
+              }}</label>
+              <div class="description">
+                {{ $t('dialog:pageSetup.lyricsMelismaCutoffWidthDescription') }}
+              </div>
+              <InputUnit
+                v-model="form.lyricsMelismaCutoffWidth"
+                class="unit-input"
+                unit="pt"
+                :min="0"
+                :step="1"
+                :precision="0"
+              />
+            </div>
+            <div class="form-group">
+              <div class="name">
+                {{
+                  $t('dialog:pageSetup.ignorePunctuationWhenPositioningLyrics')
+                }}
+              </div>
+              <input
+                id="page-setup-dialog-ignore-punctuation-when-positioning-lyrics"
+                v-model="form.ignorePunctuationWhenPositioningLyrics"
+                type="checkbox"
+              />
+              <label
+                for="page-setup-dialog-ignore-punctuation-when-positioning-lyrics"
+                >{{
+                  $t(
+                    'dialog:pageSetup.ignorePunctuationWhenPositioningLyricsDescription',
+                  )
+                }}</label
+              >
             </div>
             <div ref="textBoxesRef" class="subheader">
               {{ $t('dialog:pageSetup.textBoxes') }}
--- a/src/i18n/en/dialog.json
+++ b/src/i18n/en/dialog.json
@@ -113,6 +113,8 @@
     "melkiteRtlDescription": "Displays the score from right to left. Intended for Melkite scores in Arabic. Uses a specialized RTL neume font.",
     "lyricsMelismaCutoffWidth": "Melisma Cutoff",
     "lyricsMelismaCutoffWidthDescription": "Melisma lines shorter than this value will not be displayed. This helps prevent them from being mistaken for punctuation marks, such as periods or commas. To disable melisma lines entirely, set this to a large number.",
+    "ignorePunctuationWhenPositioningLyrics": "Ignore Punctuation When Positioning Lyrics",
+    "ignorePunctuationWhenPositioningLyricsDescription": "When enabled, leading and trailing punctuation marks (such as quotes, commas, and periods) are ignored when positioning lyrics beneath neumes. This improves visual alignment by centering the sung syllable rather than surrounding punctuation.",
     "disableGreekMelismata": "Disable Greek Melismata",
     "disableGreekMelismataDescription": "Prevents automatic use of the Greek melismatic convention, even when Greek text is detected. Useful for languages that use the Greek script but follow different notational conventions.",
     "defaultStyling": "Default Styling",
--- a/src/models/PageSetup.ts
+++ b/src/models/PageSetup.ts
@@ -199,6 +199,7 @@ export class PageSetup {
   public noFthoraRestrictions: boolean = false;
   public disableGreekMelismata: boolean = false;
   public alignIsonIndicators: boolean = true;
+  public ignorePunctuationWhenPositioningLyrics: boolean = true;
 
   public useOptionalDiatonicFthoras: boolean = false;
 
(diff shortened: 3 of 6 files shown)


basil:
@danielgarthur This appears to make the word extension too long for left-aligned lyrics:

image

basil:
@danielgarthur I think the `alignLeft` rendering path in `TheEditor.vue` needs to honor `lyricsHorizontalOffset` as a physical shift, not as padding.

For `alignLeft`, instead of forcing `left: 0`, perhaps we could honor the negative offset:

```ts
left: element.alignLeft ? Math.min(0, element.lyricsHorizontalOffset) : undefined,
```

...and skip `paddingLeft`/`paddingRight` when `alignLeft` is true (the padding semantics from the centered case don't apply, since in left-aligned mode, the offset should physically shift the box, not pad inside it). That puts the lyric box at `left: h` for leading punctuation (quote hangs left of neume), and at `left: 0` for trailing punctuation.

danielgarthur:
> @danielgarthur This appears to make the word extension too long for left-aligned lyrics:
> 
> image

In this case, should the quote be ignored for positioning and the T aligned to the left side of the neume?

danielgarthur:
Because my first inclination was just to always consider punctuation for left-aligned elements.

basil:
> In this case, should the quote be ignored for positioning and the T aligned to the left side of the neume?

It's somewhat of a subjective question (see [microtype](https://ctan.org/pkg/microtype)) but my initial preference was indeed for the quote to be ignored for positioning and the T aligned to the left side of the neume, `microtype`-style. After I wrote the above, I checked Dorico and its defaults seem to agree with me:

image

danielgarthur:
@basil 
How about this 34f53de?

I tried getting rid of padding as suggested, but getting rid of padding left definitely breaks the layout. I think padding right may be okay to nix, but I haven't test it thoroughly enough to be confident and I don't think it hurts to leave it in there.

From one of your doxologies, this is what happens when I remove padding left from left-aligned neumes. See how "hast shown" becomes too close.
image

Perhaps it's not just a matter of checking whether the element is left-aligned, but also whether the offset is positive or non-negative.

basil:
That makes sense. Commit 34f53de fixes the leading-punctuation side, but there's a symmetric case on the trailing-punctuation side that needs the same kind of accounting. I reproduced it with the lyric `"Lord.."` and saw:

image

Here's the AI analysis and fix. For a trailing-only example, let:

- `L = w("Lord")`
- `C = w(",")`
- `n = neume width`

With punctuation-aware positioning enabled, the feature sets `lyricsHorizontalOffset = +C` so the centered lyric shifts right and the syllable body, not the syllable-plus-comma, centers under the neume. For this case `lyricsLeadingPunctuationWidth = 0`.

The semantic trigger for `alignLeft` should be: "the syllable body is wider than the neume," i.e. `L > n`. If `L <= n`, the syllable fits under the neume and the trailing punctuation can hang over the right edge.

The problematic `shouldAlignLeft` check was:

```ts
noteElement.lyricsWidth - noteElement.lyricsLeadingPunctuationWidth >
  noteElement.neumeWidth - noteElement.lyricsHorizontalOffset
```

Substituting the trailing-only case:

```text
(L + C) - 0 > n - (+C)
L + 2C > n
```

So `alignLeft` fired when `L > n - 2C`, not when `L > n`. That left a `2C`-wide window where the syllable body fit but `alignLeft` still triggered.

There are two pieces to closing that:

- The `alignLeft` trigger should compare the bare syllable width directly to the neume width.
- When `alignLeft` legitimately fires, trailing punctuation should not become left padding.

The trigger-side fix is to compare the bare syllable width directly to the neume width:

```ts
noteElement.lyricsWidth -
  noteElement.lyricsLeadingPunctuationWidth -
  noteElement.lyricsTrailingPunctuationWidth >
  noteElement.neumeWidth
```

That expresses the intent directly and makes the threshold independent of which side the punctuation sits on.

The renderer-side fix is to track trailing punctuation separately as `lyricsTrailingPunctuationWidth`, then subtract that component from positive `paddingLeft` only when `alignLeft` is true:

```ts
paddingLeft:
  (!element.isFullMelisma ||
    (element.melismaText && !element.melismaText.endsWith(TATWEEL))) &&
  element.lyricsHorizontalOffset > 0
    ? withZoom(
        element.alignLeft
          ? element.lyricsHorizontalOffset -
            element.lyricsTrailingPunctuationWidth
          : element.lyricsHorizontalOffset,
      )
    : undefined,
```

For the trailing-only case, `paddingLeft = C - C = 0`, so the content is no longer shoved inward by the comma width. For mixed cases, only the trailing-punctuation contribution is cancelled and the rest of the offset remains in effect.

The melisma start calculation also needs the same trailing-punctuation awareness so the melisma span starts after the visible lyric body rather than after the ignored punctuation:

```ts
diff --git a/src/components/TheEditor.vue b/src/components/TheEditor.vue
index 86ae7e52..d09efc41 100644
--- a/src/components/TheEditor.vue
+++ b/src/components/TheEditor.vue
@@ -1382,7 +1382,12 @@ export default defineComponent({
           (!element.isFullMelisma ||
             (element.melismaText && !element.melismaText.endsWith(TATWEEL))) &&
           element.lyricsHorizontalOffset > 0
-            ? withZoom(element.lyricsHorizontalOffset)
+            ? withZoom(
+                element.alignLeft
+                  ? element.lyricsHorizontalOffset -
+                      element.lyricsTrailingPunctuationWidth
+                  : element.lyricsHorizontalOffset,
+              )
             : undefined,
         paddingRight:
           (!element.isFullMelisma ||
diff --git a/src/models/Element.ts b/src/models/Element.ts
index dc927b34..656348bc 100644
--- a/src/models/Element.ts
+++ b/src/models/Element.ts
@@ -438,6 +438,7 @@ export class NoteElement extends ScoreElement {
   public neumeWidth: number = 0;
   public lyricsWidth: number = 0;
   public lyricsLeadingPunctuationWidth: number = 0;
+  public lyricsTrailingPunctuationWidth: number = 0;
   public alignLeft: boolean = false;
   public noteIndicatorNeume: NoteIndicator | null = null;
   public scaleNotes: ScaleNote[] = [];
diff --git a/src/services/LayoutService.ts b/src/services/LayoutService.ts
index 76c47603..df9d9a1c 100644
--- a/src/services/LayoutService.ts
+++ b/src/services/LayoutService.ts
@@ -1656,14 +1656,18 @@ export class LayoutService {
     nextNoteElement: NoteElement | null,
   ): boolean {
     // At the start of a melisma, the syllable is aligned to the
-    // left of the neume, but only if the lyrics are wider than the neume.
+    // left of the neume, but only if the bare syllable (excluding any
+    // leading/trailing punctuation that is ignored when positioning
+    // lyrics) is wider than the neume.
     // NOTE: a syllable ending with a hyphen is only considered a melismatic note
     // if the next note is purely melismatic (i.e. the next note contains only a hyphen),
     // despite the unfortunate property name "isMelisma" being true.
     return (
       noteElement.isMelismaStart &&
-      noteElement.lyricsWidth - noteElement.lyricsLeadingPunctuationWidth >
-        noteElement.neumeWidth - noteElement.lyricsHorizontalOffset &&
+      noteElement.lyricsWidth -
+        noteElement.lyricsLeadingPunctuationWidth -
+        noteElement.lyricsTrailingPunctuationWidth >
+        noteElement.neumeWidth &&
       (!noteElement.isHyphen ||
         (nextNoteElement != null &&
           nextNoteElement.isMelisma &&
@@ -2812,6 +2816,8 @@ export class LayoutService {
     );
 
     noteElement.lyricsHorizontalOffset = 0;
+    noteElement.lyricsLeadingPunctuationWidth = 0;
+    noteElement.lyricsTrailingPunctuationWidth = 0;
 
     if (noteElement.lyrics.length > 0) {
       noteElement.lyricsWidth = this.getTextWidthFromCache(
@@ -2833,8 +2839,8 @@ export class LayoutService {
         noteElement.lyricsHorizontalOffset -=
           noteElement.lyricsLeadingPunctuationWidth;
 
-        // Adjust for trainling punctuation
-        noteElement.lyricsHorizontalOffset += this.getTextWidthFromCache(
+        // Adjust for trailing punctuation
+        noteElement.lyricsTrailingPunctuationWidth = this.getTextWidthFromCache(
           textWidthCache,
           noteElement,
           pageSetup,
@@ -2842,6 +2848,9 @@ export class LayoutService {
           false,
           true,
         );
+
+        noteElement.lyricsHorizontalOffset +=
+          noteElement.lyricsTrailingPunctuationWidth;
       }
     } else {
       noteElement.lyricsWidth = 0;
@@ -3222,12 +3231,14 @@ export class LayoutService {
               if (!pageSetup.melkiteRtl) {
                 start =
                   element.x +
-                  element.lyricsHorizontalOffset +
+                  element.lyricsHorizontalOffset -
+                  element.lyricsTrailingPunctuationWidth +
                   element.lyricsWidth;
               } else {
                 start =
                   element.x -
                   element.lyricsHorizontalOffset +
+                  element.lyricsTrailingPunctuationWidth +
                   element.lyricsWidth;
               }
             } else if (element.lyricsWidth > element.neumeWidth) {
```

PR #2160, opened 2026-05-19
feat: Ignore punctuation when positioning lyrics

danielgarthur:
Fixing #2154 

image


danielgarthur:
There are several other places where `alignLeft` is used as a condition. I still need to investigate whether the trailing space should be subtracted in those cases, too. If so, perhaps a cleaner solution is to not alter the lyrics horizontal offset with punctuation widths until after `alignLeft` is calculated. In this case, trailing width would be ignored if `alignLeft` is true, otherwise both widths are used.

basil:
That solution does indeed sound cleaner. I tested the AI implementation of it in commit 695dbea6 and it seems to work well on my scores as well as the examples you gave above.

PR #2161, opened 2026-05-19
fix: page break on right-aligned martyria

### LINKED ISSUE #2159: page break not working now

danielgarthur:
Fixes #2159

### DIFF
--- a/src/services/LayoutService.ts
+++ b/src/services/LayoutService.ts
@@ -952,6 +952,10 @@ export class LayoutService {
             },
           );
 
+          // Must run even when lineBreak is already true (from pageBreak or
+          // an explicit lineBreak): endParagraph reads lineBreakType to pick
+          // the finishing-glue stretch, and Left's MAX_COST stretch would
+          // compete with rightMartyriaGlue and strand the martyria mid-line.
           if (martyriaElement.alignRight) {
             lineBreak = true;
             lineBreakType = LineBreakType.Justify;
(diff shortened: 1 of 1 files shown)


danielgarthur:
@basil Would you please review this to make sure I'm not doing something stupid. This commit seems to restore the previous behavior (prior to Knuth-Plass). Namely

- Paragraph breaks on right aligned martyria should have no effect.
- Page breaks should break the page after the martyria but should not affect the position of the martyria.

src/services/LayoutService.ts
```diff
@@ -952,7 +952,7 @@ export class LayoutService {
             },
           );
 
-          if (martyriaElement.alignRight && !lineBreak) {
+          if (martyriaElement.alignRight) {
```
basil, line 959:
```suggestion
          // Must run even when lineBreak is already true (from pageBreak or
          // an explicit lineBreak): endParagraph reads lineBreakType to pick
          // the finishing-glue stretch, and Left's MAX_COST stretch would
          // compete with rightMartyriaGlue and strand the martyria mid-line.
          if (martyriaElement.alignRight) {
```

basil (dismissed):
Sorry for the regression and thanks for fixing this. I know I tested page breaks at one point, but it was literally years ago and the branch had evolved significantly since then (and I obviously don't use this feature in my own scores). Noted for the future as an area to do manual testing.

PR #2172, opened 2026-05-20
Localize CKEditor UI language

basil:
The CKEditor 5 rich-text editors used for text boxes and annotations always rendered their UI (toolbar tooltips, menus, dialogs) in English, regardless of the user's chosen application language. This change makes the editor UI follow the app's language preference.

- **Bundle CKEditor translation packs** (`src/customEditor.ts`): import and register the `el`, `id`, and `ro` translation packs via the `translations` config array. English is built in and needs no pack -- together these cover every language the app supports (`el`, `en`, `id`, `ro`). The `defaultConfig.language` is split into explicit `ui`/`content` keys.
- **Resolve the editor UI language** (`src/components/TheEditor.vue`): add a `ckeditorLanguage` computed property that derives the UI language from the user's preference (`resolveLanguagePreference(editorPreferences.language, navigator.language)`), falling back to `en`. It is threaded down to all `<TextBoxRich>` and `<Annotation>` instances via a new `editor-language` prop.
- **Apply the language per editor** (`src/components/TextBoxRich.vue`, `src/components/TextAnnotation.vue`): accept the `editorLanguage` prop and pass it as `language.ui` in the per-instance editor config. Because CKEditor fixes its UI language at construction time, each `<ckeditor>` element is given a `:key` derived from `editorLanguage` so it remounts when the preference changes.

## Testing

Switched the application language to Greek, Indonesian, and Romanian and confirmed the rich-text editor toolbar/tooltips appeared in that language.

### DIFF
--- a/src/components/TextAnnotation.vue
+++ b/src/components/TextAnnotation.vue
@@ -8,6 +8,7 @@
     @dblclick="handleDoubleClick"
   >
     <ckeditor
+      :key="editorLanguage"
       ref="editor"
       class="rich-text-editor"
       :editor="editor"
@@ -47,6 +48,10 @@ export default defineComponent({
     fonts: {
       type: Array as PropType<string[]>,
       required: true,
+    },
+    editorLanguage: {
+      type: String,
+      default: 'en',
     },
     selected: Boolean,
   },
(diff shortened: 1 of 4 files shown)


basil:
Closing in favor of #2279

Issue #2034, opened 2026-03-30
Barlines should be centered between neumes


basil:
This looks off:

<img width="1280" height="610" alt="Image" src="https://github.com/user-attachments/assets/c2598ec3-836b-4038-93be-db7e3ab56cd4" />

The barline should be centered between the two neumes.

danielgarthur:
My thought here would be to use `position: absolute` on the bar lines and calculate the offset after lines have been positioned.

basil:
Possibly, but I think it's more subtle than just repositioning them. Consider this:

<img width="3954" height="694" alt="Image" src="https://github.com/user-attachments/assets/85534a8e-8017-4866-bf37-4c96da25e7c5" />

Here the last barline is actually creating uneven spacing, because it could fit completely in the glue. So I don't think this is just a matter of repositioning them, but rather perhaps they should be moved out of the neume box completely. Then Knuth-Plass could do something fancy with them, like give them stretchability on both sides, and not add `neumeDefaultSpacing` around them if it is already covered by the measure bar's width itself.

basil:
We probably need to have a broader discussion about the future of measure bars, given issues like https://github.com/neanes/neanes/issues/151#issuecomment-4137989180 and the like. If measure bars don't make sense as part of a neume-box from a spacing perspective, then they might not make sense as part of a note element from an object model or UI perspective either, except in specific cases like kentimata below oligon or yporroe.

It might make sense for them to be their own first class objects in the object model and UI, with various logic to display them appropriately. For example, if an abstract half-height measure bar is on the stream of note elements, then inserting a martyria before it should display the bar above the martyria, and deleting the martyria should promote it back. The user shouldn't have to concern themselves with such implementation details when entering the music; they should just enter barlines on an abstract level as they relate to the music, and the software should internally display or position them in the right place.

basil:
I wrote about my ideas about "grouping" notes together as a first class UI concept in the other thread. I think if something like that were implemented on the frontend, then it would be straightforward for me to update the (Knuth-Plass based) layout service to do the right thing here and lay it all out correctly. Unfortunately I am probably not the best person to do major UI features, even with AI assistance.

danielgarthur:
Closed by #2209

PR #2190, opened 2026-05-23
fix: Add space after yporroe

danielgarthur:
With neume spacing set to the default 0.03 in:
image

With neume spacing set to 0:
image


danielgarthur:
@basil Would you please take a look at your scores with this font change? 

My only concern is that the default Neanes neume spacing setting makes these perhaps too far apart. If the neume spacing setting is set to 0, then the result is as seen in the examples from Fr. Ephraim that you provided.

But I wonder if that spacing is even necessary anymore since your Knuth-Plass change? I checked a couple of your scores and set the spacing to 0 and it seems alright. Maybe the new default neume spacing should be 0?

basil:
I agree that ~2.1 points of spacing is on the generous side, but I'd be nervous about setting the default any lower than 1 point. Below that, adjustment ratios start to get wildly high, since there isn't enough fixed-width material to work with to fill each line, and I suspect it drowns out the effect of discouraged breaks. I had it in the back of my mind to decrease the value in my own scores to 1.5 points, which (given the 50% stretch) would effectively put an upper bound at the old value of ~2.1 points while trying to aim for lower if possible. I haven't pulled the trigger on that yet in my own scores because I wanted to check over the effects on the statistics first, but my general sense is that it should be fine, and it would probably give the scores a more polished and tight appearance. The only reason people usually want very loosely typeset Byzantine music is to prevent adjacent lines from looking too different when one of them has rivers of whitespace caused by bad line breaking, and (as you noted) that problem is essentially solved by Knuth-Plass.

---

I looked at a few of my scores with this change. The biggest wins were cases where a barline immediately followed, as these almost always resulted in the gorgon crashing into the barline and requiring manual adjustment. These now look great:

image

Even in cases where a collision wasn't imminent, I still see an improvement, such as this line, where the digorgon would have otherwise gotten a little too close to the next klasma for comfort:

image

However, in the middle of a melisma, without any barlines, things look worse:

image

Here's an interesting case, where the digorgon looks about right, but the first gorgon is still too close, even with these changes:

image

I'm not sure offhand what the best way to deal with this is. This almost feels like a kerning issue. Normally kerning isn't a problem in Byzantine music because there aren't very many opportunities for notes to "tuck under" each other in a pleasing (or not pleasing) way, but this seems like one of them. I think the traditional solution to this type of problem is to build a table of kerning pairs. Another related question I have is how are you determining the bounding box of each glyph, because it feels like they are inconsistent, with some of them stretching slightly outside the bounding box (like the last example above) and sometimes the bounding box extending past the end of the glyph (and thus inflating the spacing between neumes). If all the bounding boxes were perfectly accurate then maybe 2 points of space between each neume wouldn't be so bad?

danielgarthur:
To answer the question about the bearings, they were taken from the original EZ Psaltica font. So they have peculiarities based on whatever looked good in Word plus what was possible with the VB macros. I did at one time experiment with a "tight" version of the font that left no space to the left and right, but it did not look good.

For the running elafron with the petaste, the outline is outside the bearings in the original font, too. 

For the melisma, it looks better with 1pt of neume spacing, but beyond that, defining kerning pairs would be necessary, as you said.

basil:
> I did at one time experiment with a "tight" version of the font that left no space to the left and right, but it did not look good.

I wonder if it would look better now that we have Knuth-Plass. If you look at classical Byzantine music books from the 1800s, most of them are a lot tighter, as are scores produced by Melodos (one of the few things I actually like about Melodos scores). Not sure how much of that is due to the bearings or the default spacing between neumes, but even Elias was trying to set the spacing between neumes to a negative (!) value, so I think that is evidence from users that the bearings are a little too wide.

> For the running elafron with the petaste, the outline is outside the bearings in the original font, too.

Possibly a bug? I haven't dived too deeply into the technical aspects of building a font, but it seems strange to me intuitively that anything would ever be outside its bearings. Kerning, yes (that is the point), but bearings seem like they would be mostly (or entirely) positive?

danielgarthur:
Would it help if there were a separate input for the max stretch/shrink and the neume spacing were defaulted to 0?

Ultimately the neume spacing parameter was some kind of compensation for other issues that may not even exist anymore. I think it was mainly to compensate for some lyric issues that were addressed a long time ago. I would prefer it if all default spacing were encoded in the font. This may involve updating the font kernings a little more.

basil:
> Would it help if there were a separate input for the max stretch/shrink and the neume spacing were defaulted to 0?

Something like this could work. My main concern is to keep Knuth-Plass working with standard operating parameters, that is, almost all lines should have an adjustment ratio between -1 and 1, ideally 0, with a small number of lines having an adjustment ratio between 1 and 1.5, and a very small (pathological) number of lines having an adjustment ratio > 2. Currently, setting neume spacing to 0 pushes the adjustment ratio of many lines over 10, and I suspect this outcompetes the discouraged breaks. But I am open to changing the defaults or the Knuth-Plass implementation to make spacing tighter (and think this is a good idea), just as long as Knuth-Plass continues to work with standard operating parameters (reasonable adjustment ratios).

> I would prefer it if all default spacing were encoded in the font. This may involve updating the font kernings a little more.

Agreed, the font seems like the right place for this metadata.

danielgarthur:
My thoughts on the "tight" font (no space between the contours and the side bearings) are that it might work, but not with a single "standard glue". There would have to be potentially as many glues as there are neume pairings. For example, in EZ Psaltica, the space between apostrofos-apostrofos is much smaller than the space between ison-ison.

These pairwise standard glue widths should be easily calculable via a Python font forge script from the current EZ Psaltica font.

Somewhere I probably still have the script that eliminates all the space in the font. If not, I'm sure AI could write it nowadays.

basil:
That makes sense, and also sounds like a much better architecture for tight scores compared to setting the neume spacing to 0 in the current code. In the current code, setting neume spacing to 0 relies on the existing bearings but messes up Knuth-Plass with enormous adjustment ratios. With the approach you're describing, the pairwise standard glue widths would be the "material" that Knuth-Plass needs to work with, adjustment ratios wouldn't blow up, and a tight score could be achieved without outcompeting the logic for discouraged breaks. And on top of that we could layer on the existing neume spacing setting (how many points of additional space to add or subtract from the pairwise standard glue widths), and optionally additional setting(s) for how much the glue can stretch or shrink (currently hard-coded to 50%).

danielgarthur:
It turns out the script that shrinks the bearings is still in the sbmufl repo. One of the first challenges that comes to mind is that the vareia and the bar lines are part of a single Neanes "syllable" element, which is currently positioned as a single box. The algorithm would need to split vareia + ison into two "virtual elements" with their own boxes and glues. The result would then be position for the syllable element, as well as an internal position for the vareia. Or at least that is one way of going about it. 

Another way would be to create a separate vareia element that is treated like any other element except that it wouldn't be selectable in the UI. Pressing the "vareia" button an a neume would add or remove this element. I lean towards this idea.

If you have other ideas, please let me know.

danielgarthur:
Although maybe the first idea would actually be easier... I'm not sure. I would have to play around with it and see.

basil:
I'm all in favor of splitting the vareias and measure bars into their own elements for layout purposes. See https://github.com/neanes/neanes/issues/2034#issuecomment-4203174454 and https://github.com/neanes/neanes/issues/151#issuecomment-4209724064. One of my biggest remaining issues with my current scores is that measure bars aren't centered between neumes and that they disrupt spacing. Since measure bars are mostly white space in the main flow of reading, they can essentially be "folded in" to the spacing between neumes without adding many points of additional space like they do today (which, on top of the already generous 2 point default spacing between neumes, feels excessive). So I think separating them out into their own elements would be a good first step in this direction.

danielgarthur:
Closing in favor of #2209. Regardless of whether #2209 lands, this change is probably not going to work because of the spacing issues.

Issue #2178, opened 2026-05-21
Knuth Plass phase 2


basil:
Now that Knuth Plass phase 1 is finished, here are my thoughts on phase 2. I probably won't have the time to work on this soon, but jotting down my thoughts while they are fresh in my mind.

The main thing that is needed to beat the classical publications in quality is an implementation of these techniques:

> The 19th-century publications also use several techniques to improve the quality of first-fit output:
> 
> 1. To compress a line, reduced-width versions of neumes such as the ison and oligon are sometimes used.
> 2. To compress a line, an ison followed by an apostrophos is sometimes written with the apostrophos beneath the ison rather than after it.
> 3. To compress a line, two apostrophoi are sometimes written with the second apostrophos beneath the first rather than after it.
> 4. When a line is too narrow, extra space is often added first to the martyriae and only then to the other neumes, so that the spacing of the ordinary neumes stays more consistent with the surrounding lines.

We currently implement 4 using stretch, which works to some degree when the line is already getting stretched (within the adjustment ratio). A more complete implementation would allow martyria to stretch even when the rest of the line is still at a normal adjustment ratio. That, and the first three points, require something like the idea Knuth had for typesetting biblical Hebrew, which is never hyphenated:

> Hebrew fonts intended for sacred texts usually include wide variants of several letters, so that individual characters on a line can be replaced by their wider counterparts in order to avoid wide spaces between words. For example, there is a super-extended aleph in addition to the normal one.

Knuth & Plass acknowledge that "an extension would be needed to make the optimum-fit algorithm handle cases like [these]. ... The badness function of a line would [...] depend not only on its natural width, stretchability, and shrinkability," but also on "the number of dual-width characters present."

I am not aware of anyone who has implemented such an extension, but conceptually it is straightforward enough. Implementing it efficiently is another story: this algorithm is well known for its quadratic time explosion if not carefully optimized.

The way I envision this working is as follows: the smaller-width ison from the "ison and apostrophos" glyph should be extracted to its own glyph, and a similar sized oligon glyph should be created as well. The existing "double apostrophos" and "ison and apostrophos" glyphs should be removed from the UI: there is no reason to ever create these manually, as they exist for typographical reasons, not for semantic reasons. A new checkbox should be added to the settings (on by default) for allowing the use of these replacements ("ison and apostrophos", "double apostrophos", "small ison", and "small oligon").

[...]

PR #2209, opened 2026-06-02
Switch to tight font

### LINKED ISSUE #862: Melisma spacing too tight
In https://github.com/basil/byzantine-music/blob/1c7680a54a9c93c8c91ff557f5d6e1d27a59aff3/Mysteries%20and%20Sacraments/Paraklesis/18%20Ode%20Seven%20-%20Small.byzx the melisma spacing is too tight here:

spacing

While this has been tolerable in some other cases, in this case I think it has a negative effect on readability. I don't think a lyric syllable should ever extend beneath a preceding melisma neume, as "womb" does in this example. I suppose the expected behavior would be for an lyric extender (underscore) to be drawn for the "Maid's" melisma, with some extra whitespace added to the left of the apostrophos for "womb" (just as it already is to the right of the apostrophos).

### LINKED ISSUE #2183: Add space after yporroe
Papa Ephraim's original macros contained 100 lines of code to add space after an yporroe that has a gorgon, digorgon, or trigorgon above it, regardless if there is an aple or diple beneath them:

Image

This has regressed in Neanes:

Image

This causes frequent problems in my scores at the ends of lines with barlines:

Image

danielgarthur:
## Summary
- Add glyph-specific leading and trailing spacing metadata for supported fonts.
- Replace uniform inter-neume spacing with glyph-aware boundary glue plus the user-configurable neumeDefaultSpacing adjustment.
- Resolve note-local contextual substitutions before width and spacing lookup.
- Add vareia internal spacing and RTL-aware rendering.
- Apply glyph-aware spacing around notes, martyriae, tempo signs, text boxes, images, and drop caps.
- Center barlines between neighboring elements and preserve metadata-defined minimum spacing.
- Add focused layout regressions for substitutions, vareia, inline content, and barline positioning.

## Known Issues

**Fixed**
~~The leading spacing for martyria is large in the original EZ Psaltica font and probably needs to be reduced.
image~~

~~Right bar lines attached to martyria do not center between the martyria and the neume.~~
image

~~Putting the bar line to the left of the next neume works, though.~~
image~~


danielgarthur:
@basil Here is my first attempt at vibe coding... Is it hot garbage or workable?

I'm not sure at which point AI is just leading me down a rabbit hole where every time I ask it to review for "feature completeness" it always finds something...

It seemed to mostly work, with the exception of the two issues mentioned in the PR description. For the first issue, I think the zero-width font needs to eliminate the extra leading space. Perhaps making it equal to the trailing space. That extra leading spacing is presumably and artifact of convenience from back when it was used in Word.

You can see all the extra leading space in Font Forge for a martyria note here: 
image


~~For the second issue, I burned through so many tokens, I gave up. I think I could probably figure it out just by debugging it myself. Not sure if that should be fixed now or later.~~ Fixed with just a little bit of human thought.

danielgarthur:
@basil  I also suspect that `spacingRamp` needs to be fixed, but I'm not sure how.

basil:
This definitely seems workable.

> I'm not sure at which point AI is just leading me down a rabbit hole where every time I ask it to review for "feature completeness" it always finds something...

Yes, this is a known problematic behavior. I encounter it when I ask if the code needs to be cleaned up any further, which it interprets as a polite request to do more refactoring, even well past the point of the code being already refactored and clean. To combat this obsequiousness I use https://github.com/basil/dotfiles/blob/2aed14bdf1bdb627e9f7b4139f42804eaa242081/ansible/roles/dot-base/files/home/user/.agents/skills/stop-yes-manning/SKILL.md

> It seemed to mostly work, with the exception of the two issues mentioned in the PR description.

I noticed two other things: With these changes the "oligon with kentima" neume has extra space between the oligon and kentima, which seems wrong. Martyrias on the high range (with the apostrophe) also don't display properly yet. 

> For the first issue, I think the zero-width font needs to eliminate the extra leading space. Perhaps making it equal to the trailing space. That extra leading spacing is presumably and artifact of convenience from back when it was used in Word.

Makes sense, we may never get an answer to https://github.com/neanes/sbmufl/discussions/22 but I think we can move on by eliminating the extra leading space.

> I also suspect that spacingRamp needs to be fixed, but I'm not sure how.

We were computing the ramp's high anchor as `neumeDefaultSpacing`. That was 2.88px (0.03") when written; this PR made it 0. The denominator flipped from +2.75 to -0.13, so the ramp became inverted. I think we could instead key on the actual gap determined by the bearings + `neumeDefaultSpacing` (0 by default). Then normal neumes would land at about k = 0.2 at the default, roughly preserving the old behavior. This matches the Knuth Plass doc's stated intent ("does not dominate when the base spacing is already small"). Either way, notes/knuth_plass.md:161-165 and the LayoutService.ts:478 comment need updating ("(approximately 2.16 pt)" is no longer the default).

If `spacingRamp` is too much trouble, you can delete it and just space around martyrias the same as any other pair of neumes. I'm not happy with how it turned out, and the attempts to add a ramp were more of a short-term hack. The long term fix is described in https://github.com/neanes/neanes/issues/2178.

basil:
At a high level, this change is really great, and my scores look fantastic at 0 spacing with the barlines nicely centered between neumes. This is a very impressive change.

One remaining issue I still see with barlines is the one described in https://github.com/neanes/neanes/issues/2034#issuecomment-4157616221, where a neume that has a syllable longer than the neume, and which appears at either the beginning or end of a line, has the barline stuck to the neume. There isn't really precedent for this in Greek scores, because it's very rare in a Greek score for a syllable to be large than its surrounding neume. But in English this happens frequently, and my gut feeling is that the bar line should always be flush with the left- or right- margin of the page in those cases.

My subagent also tells me that in `centerMeasureBars` the terminal-right -spacing render branch can miss a narrow edge case (right barline as last visible anchor with a non-anchor element after it on the same line). Perhaps worth testing visually.

danielgarthur:
@basil Would you please send me pictures of these two issues? I do recall there being issues with those, but I thought they were fixed.

> I noticed two other things: With these changes the "oligon with kentima" neume has extra space between the oligon and kentima, which seems wrong. Martyrias on the high range (with the apostrophe) also don't display properly yet.

image

basil:
> @basil Would you please send me pictures of these two issues? I do recall there being issues with those, but I thought they were fixed.

It's fixed as of https://github.com/neanes/neanes/pull/2209/commits/878d27432dad609a5effb64fa94409d445d9658d, but when I rebuild SBMuFL at https://github.com/neanes/sbmufl/pull/49/changes/f9f4529bb998576c40e1998909473ded7a248364 the issue comes back, so perhaps you never commited the fix to https://github.com/neanes/sbmufl/pull/49.

danielgarthur:
> > @basil Would you please send me pictures of these two issues? I do recall there being issues with those, but I thought they were fixed.
> 
> It's fixed as of [878d274](https://github.com/neanes/neanes/commit/878d27432dad609a5effb64fa94409d445d9658d), but when I rebuild SBMuFL at [neanes/sbmufl@f9f4529](https://github.com/neanes/sbmufl/commit/f9f4529bb998576c40e1998909473ded7a248364) the issue comes back, so perhaps you never commited the fix to [neanes/sbmufl#49](https://github.com/neanes/sbmufl/pull/49).

I see the fix in PR, so I'm not sure...

basil:
A somewhat subjective question I have is, should the shrink be equal to the stretch when the neume default spacing is 0? I worry that at the lowest adjustment ratio (-1) the shrink might be too tight and not look good, since we're effectively blowing through the bearings in this case. TeX's normal interword space is 3.333pt plus 1.667pt minus 1.111pt, stretch = ½, shrink = ⅓ (ratio of 3:2). So maybe we could or should do something similar here. This was less of a problem with the old code because we were never down to the wire when shrinking. Might be worth inspecting some lines with "show adjustment ratios" enabled to see if such a change makes the ones with -1 adjustment ratios look better.

basil:
I haven't looked at the code too closely, but on my first pass it looked like it had grown organically by accretion: lots of near-duplicate helpers, mirror-image blocks written out twice, etc. My AI generated code generally looks the same and I usually do a few rounds of cleanup after everything is working. The "hacky vs principled" prompt wording in https://github.com/basil/dotfiles/blob/2aed14bdf1bdb627e9f7b4139f42804eaa242081/ansible/roles/dot-base/files/home/user/.agents/skills/subagent-review/SKILL.md is generally useful, although I have to be careful to prevent obsequiousness from turning a cleanup request into an infinite rabbit hole of refactoring.

basil:
Additionally there seems to be a bug in this scenario: a short neume at the end of a line (like apostrophos) with a lyric that is larger than the neume. When that last neume does not have a right barline, or when its right barline is physically attached to that neume, the lyric is properly centered. But when that last neume gets an inferred right barline because the next line's first neume has a left barline, then the lyric goes off center. :grimacing:

danielgarthur:
> Additionally there seems to be a bug in this scenario: a short neume at the end of a line (like apostrophos) with a lyric that is larger than the neume. When that last neume does not have a right barline, or when its right barline is physically attached to that neume, the lyric is properly centered. But when that last neume gets an inferred right barline because the next line's first neume has a left barline, then the lyric goes off center. 😬

I think this is a pre-existing bug. Although I may take a look at it anyway with this PR if it doesn't add too much.

danielgarthur:
> At a high level, this change is really great, and my scores look fantastic at 0 spacing with the barlines nicely centered between neumes. This is a very impressive change.
> 
> One remaining issue I still see with barlines is the one described in [#2034 (comment)](https://github.com/neanes/neanes/issues/2034#issuecomment-4157616221), where a neume that has a syllable longer than the neume, and which appears at either the beginning or end of a line, has the barline stuck to the neume. There isn't really precedent for this in Greek scores, because it's very rare in a Greek score for a syllable to be large than its surrounding neume. But in English this happens frequently, and my gut feeling is that the bar line should always be flush with the left- or right- margin of the page in those cases.

I'll look into it.

> 
> My subagent also tells me that in `centerMeasureBars` the terminal-right -spacing render branch can miss a narrow edge case (right barline as last visible anchor with a non-anchor element after it on the same line). Perhaps worth testing visually.

After checking the code and testing the app visually, I don't think there is any non-anchor element that could possibly appear to the right of an anchor element.

danielgarthur:
@basil I'm not sure what to make of `martyriaTrailingPadding`. It seems like it is adding extra right padding all the time. https://github.com/neanes/neanes/blob/0054abcfbef2d5cf7e8cbe686479e9c0bbe528de/src/services/LayoutService.ts#L932

Should this only happen if it's a right-aligned neume with a quantitative neume instead of all the time?

basil:
I asked myself that same question while implementing Knuth-Plass. :smile: Intuitively it felt like the extra right padding should only be added if it's a right-aligned neume with a quantitative neume. However, on further testing against older versions, I saw that the older versions were always adding it unconditionally... and that furthermore, some (but not all) martyria combinations spiilled out of their "bounding box" (not sure what the right term for this is) and that extra padding helped keep them from going into the right margin. Ultimately I was just trying to preserve existing behavior without introducing any regression... but I was equally confused about the existing behavior. To me it intuitively seemed like the "bounding box" should have been more exact, and that padding shouldn't have been needed...

basil:
While we're on the subject of space after martyria, there was a weird `public spaceAfterMartyriaFactor = 0.148` in `PageSetup.ts` that I didn't understand either, but tried to preserve anyway.

danielgarthur:
@basil At this point, the PR is getting so large, and I still haven't been able to clean it up as successfully as I want. So for now I want to focus only on fixing regressions. I'm not going to try and put the measure bar flush against the margins in that one case in this PR. I only included the measure bar centering in this PR because bar lines would have otherwise been completely broken.

One major issue that I'm seeing is this: the space between tempo and martyria is too large.

Before:
image


After:
image

This leads to disastrous results at times. For example in your Stichera of Pascha.

The result is better when using the `tempoLeft` and `tempoRight` feature of the martyria.

I'm thinking maybe the metadata needs to have an optional context aware spacing. Something like

```json
"glyphSpacings": {
  "martyriaNoteDi": {
    "leading": 0.12, // default
    "trailing": 0.12, // default
    "contexts": {
      "tempo": {
        "leading": 0.06, // override for tempos
        "trailing": 0.06 // override for tempos
      }
    }
  }
}
```

I'm also unsure what the default spacing should be when the martyria is between two neumes. I would appreciate your typographical experience here if you have suggestions.

I'm also not sure what the expectation should be here:
image
Should the martyria + tempoLeft/tempoRight combo be centered? Or should it remain as is, where the tempo sign has less leading/trailing space and is therefore closer to the neume?

danielgarthur:
Maybe for simplicity, I could instead start with a top level setting that sets the default space for all martyria-neume combos, and martyria-tempo combos. I don't know that it's worth specifying the spacings for each `martyriaNote*` option. 

```json
"engravingDefaults": {
  "spacing": {
    "martyriaToNeume": 0.18,
    "martyriaToTempo": 0.06,
  }
}
```

basil:
> The result is better when using the tempoLeft and tempoRight feature of the martyria.

Maybe we should just support only that. I can migrate my scores easily enough.

> I'm also unsure what the default spacing should be when the martyria is between two neumes. I would appreciate your typographical experience here if you have suggestions.

At the tight end, martyriae can be treated like any other neume, with similar small bearings around them like an ison or oligon. However, they can stretch a lot more, especially in the older publications (which I admire). In John's _Anastasimatarion_ and _Heirmologion_ (early 1900s) they stretch a moderate amount. In the even older books (which I admire very much) like the 1825 Heirmologion, https://analogion.com/psaltologion/neoklis/Tameion_Xourmouziou_A.pdf and https://stavropoleos.ro/wp-content/uploads/biblioteca-digitala/Psaltichie/psaltichie_carte_veche/pann_heruvicochinonicar_vol1_buc_1847_c5.pdf they can stretch quite a bit more, effectively absorbing most of the slack in a short line (up to an extent) and allowing the rest of the neumes to be typeset at normal spacing while still filling the line.

I had trouble modeling the above in Knuth-Plass. I was able to add large stretch to martyrias, but it wouldn't be taken unless adjustment ratios were already high. In other words, I could only get martyrias to stretch when all the other neumes were stretching, not what I wanted (for martyrias to stretch so that other neumes wouldn't have to stretch). There is no way in stock Knuth-Plass to "prioritize" one stretch over another. I ultimately concluded that Knuth-Plass needed yet another extension to handle this (described in #2178).

Until the above can be implemented in a future project, there should probably be spaced a "moderate" amount by default, something like

image

or

image

> Should the martyria + tempoLeft/tempoRight combo be centered? Or should it remain as is, where the tempo sign has less leading/trailing space and is therefore closer to the neume?

I will try to find some examples in classical books to be sure, but my intuition is that when a tempo sign is to the left or right of a martyria, it shouldn't have any "extra" stretch beyond the usual bearings, but the entire "group" (martyria + tempo) should be able (ideally) to stretch on both sides a significant amount as described in the preceding few paragraphs.

basil:
> Maybe for simplicity, I could instead start with a top level setting that sets the default space for all martyria-neume combos, and martyria-tempo combos. I don't know that it's worth specifying the spacings for each `martyriaNote*` option.

I don't really see a need to encode anything in the font other than the smallish bearings for each martyria, which would be the same whether it was by itself or surrounded by a tempo sign:

image

The "extra" on top of that to achieve the "moderate" spacing I described above IMHO belongs in the program, not in the font, and (I think...with 80% confidence) would apply to the entire "group" of martyria + tempo sign, not just to one or the other.

danielgarthur:
> > Maybe for simplicity, I could instead start with a top level setting that sets the default space for all martyria-neume combos, and martyria-tempo combos. I don't know that it's worth specifying the spacings for each `martyriaNote*` option.
> 
> I don't really see a need to encode anything in the font other than the smallish bearings for each martyria, which would be the same whether it was by itself or surrounded by a tempo sign:
> 
> image
> The "extra" on top of that to achieve the "moderate" spacing I described above IMHO belongs in the program, not in the font, and (I think...with 80% confidence) would apply to the entire "group" of martyria + tempo sign, not just to one or the other.

Apologies, but I'm not sure what you are proposing regarding the martyria bearings. Here is what your stichera looks like with minimum bearings.

image

It looks a little tight to me compared to the examples you provided, although when I drop the martyriae through the Page Setup settings, it looks perhaps a little closer.

image

What is your opinion?

What did you have in mind when you said that the program should be in charge of the spacing? Are you saying there should be a setting to control the space around the martyria?

Edited: 
I forgot to set the neume default spacing to 0 in the first post. I have updated the images in this post so that they use 0 spacing.

basil:
Right, what I was trying to say above was that:

- The font itself should just contain the minimum bearings, which (if not augmented by the program) would look like what you just showed above.
- The program should augment that with a generic (font agnostic) algorithm.

As to that generic (font agnostic) algorithm, I tried to divide that into two ideas:

1. The "ideal" idea (not possible today): If Knuth Plass could support a continuum of possible widths for a given glyph (or even a handful of fixed size widths for a given glyph) as described in #2178, then we could achieve the look in the 1825 Heirmologion, https://analogion.com/psaltologion/neoklis/Tameion_Xourmouziou_A.pdf and https://stavropoleos.ro/wp-content/uploads/biblioteca-digitala/Psaltichie/psaltichie_carte_veche/pann_heruvicochinonicar_vol1_buc_1847_c5.pdf (where martyria can stretch quite a bit to absorb most of the slack in a line).
2. The practical idea for today: Aim for the "moderate" spacing in the Anastasimatarion and Heirmologion of John, or the two screenshots from https://github.com/neanes/neanes/pull/2209#issuecomment-4615726087.

As to how to implement 2) today, my current code calls this the "martyria bonus" term. I don't think we need a user preference for this and I'd be reluctant to expose one and have to support it forever. The amount could just be changed to be a multiple of the bearings (like 3x) or a multiple of the `neumeDefaultFontSize` (like 20%). It should be able to shrink a lot (possibly 100% of the bonus term, all the way down to just the bearings themselves, yielding the results from https://github.com/neanes/neanes/pull/2209#issuecomment-4616562860 for adjustment ratio -1), and also to stretch a fair bit (say, 50% - 100% of the bonus term at adjustment ratio > 1). At adjustment ratio 0 the results should look close to the martyria in the Heirmologion / Anastasimatarion of John, or the images from https://github.com/neanes/neanes/pull/2209#issuecomment-4615726087.

danielgarthur:
Okay, I took a pass at it. I didn't review the code too closely, but visually it looked alright. I'll look at it more later.

basil:
This looking really great, massive improvement. https://github.com/neanes/neanes/pull/2209#issuecomment-4607716619 remains an open question. I think limiting shrink to 1/3 (instead of 1/2) of bearings, like TeX, is probably prudent. Sure, it will drive up adjustment ratios modestly, but I think we have a responsibility to respect bearings. :smile: And we're pretty competent at optimizing adjustment ratios, so this seems like a tradeoff that is worth making.

Similar thoughts apply to the new martyria bonus; being allowed to eat away the entire bonus seems mildly aggressive. Perhaps we should protect at least a third of it or something. Similarly I'm not sure if it should be allowed to stretch to 50%, 75% or 100%. I think visual inspection is the only way to pick a good value.

One thing that definitely seems like an issue is that right now the root sign isn't counted as part of the "bounding box", so for example a Neanes or a Necheanes root sign that is very wide doesn't get counted and we implicitly rely on the bonus to protect us in these cases. But in the current code even that bonus can get eaten away 100% since none of it is protected and we end up with a collision when adjustment ratios are low:

image

Ideally I think it should be included in the bounding box, but if that is prohibitively difficult to implement then we definitely need to protect some of the bonus to compensate.

basil:
This keeps looking better and better. Is there anything that can be done in the short term, short of defining kerning pairs, about the extra space after the yporroe in this example:

image

Even in Papa Ephraim's book it doesn't seem to have that much space:

image

This formula is really common so I think it's important that it look good.

danielgarthur:
Quick and dirty fix is to just reduce the width by the standard glue width. Not sure whether that puts the measure bar intolerably close. What do you think? It also means that I'll need a separate glyph for the dotted gorgon, as can be seen below.
image

basil:
Yeah, that does look better.

danielgarthur:
This PR is hopefully nearing completion. The martyria width issues and the gorgon spacing issues have now been addressed.

One notable change is that I decided to abandon the leading/trailing spacing approach. It added a fair amount of complexity and, in practice, did not seem necessary given that the Knuth–Plass algorithm already provides stretch and shrink. Instead, I settled on a fixed glue width of 2.6pt for a 20pt font (13% of the font size), based on measurements from the original EZ Psaltica font. This resulted in a much simpler and cleaner implementation.

I also implemented two spacing/anti-collision mechanisms in the LayoutService. These potentially could have been encoded in the font, but the implementation is significantly simpler and more maintainable in the layout layer.

First, I discovered that the HTML canvas APIs can measure the actual rendered bounding boxes of glyphs (including combining marks and other glyphs with no advance width) via `actualBoundingBoxLeft`/`actualBoundingBoxRight`. We now use these "ink bounds" to detect and prevent visual collisions.

For elements that contain measure bars, additional glue is automatically added whenever the rendered bounds indicate that a collision could occur. This addresses cases such as yporroi with gorgon, dotted gorgon, digorgon, and similar combinations.

Second, I introduced a mapping called `noteVisualCollisionPairs`, which contains a curated list of potentially dangerous mark combinations. For example, an antikenoma on one neume followed by a sharp accidental on the next can visually collide, especially when using the tight engraving font. When one of these known pairs is encountered, we compare the actual ink bounds and add additional glue only when necessary.

The reason for using a curated list is that raw ink-bound overlap is often too aggressive. For example, a mark positioned above the first neume and a mark positioned below the second neume may technically overlap in their combined bounding boxes even though the marks themselves are visually separated. Treating every bounding-box overlap as a collision adds unnecessary spacing and produces poorer engraving. Instead, we only perform collision checks for pairs of marks that are known to have the potential to interfere visually.

One additional detail is that collision pairs may involve either marks or entire neumes. For example, a flat accidental can collide with certain neume shapes, such as an oligon + apostrofos + kentemata. The collision detection also takes any user-applied offsets into account. If a neume has been repositioned in a way that avoids the collision, the additional glue will be reduced or eliminated as appropriate. For instance, a sharp accidental can often be tucked underneath an apostrofos, allowing the pair to fit without requiring extra spacing.

The `noteVisualCollisionPairs` mapping is intentionally conservative and is based on combinations that are known to cause problems in practice. If you encounter additional collision cases in real scores, please let me know so that they can be added to the list.

danielgarthur:
@basil I have been going through all your scores with zero-space and they mostly seem good to me, although I don't have the best eye for this sort of thing.

One question I have: currently we make sure that there is at least the standard glue width between measure bar lines and neumes. I'm wondering if that is too much by default. Perhaps there should be a separate "minimum bar line glue", or we should use a multiple of the standard glue like 1/2. Most of the time, the spacing around bar lines looks good to me, but in other places it seems like a bit much.

I'm trying not to overanalyze too much because this PR will never land if I do, and I know I have the ability to fuss over things forever. At the same time, I want to lay down a good foundation and make sure that you are satisfied with your scores.

basil:
Overall, this is a massive improvement, and I am very grateful for it. Some of the worst breaks in my scores have now been solved. I agree that we need to wrap this up in a practical way rather than letting it drag on forever, so I will focus on the things that stood out to me the most.

---

First thing I noticed is lyric tucking seems to have regressed from `master`:

image

---

Another issue is there seems to be too much space after the first barline here. Perhaps the ink bounds is getting too aggressive with the antikenoma, which wouldn't have collided with the barline anyway:

image

---

> One question I have: currently we make sure that there is at least the standard glue width between measure bar lines and neumes. I'm wondering if that is too much by default. Perhaps there should be a separate "minimum bar line glue", or we should use a multiple of the standard glue like 1/2. Most of the time, the spacing around bar lines looks good to me, but in other places it seems like a bit much.

Yes, barlines still seem too disruptive to spacing to my eye, especially the middle barlines, but even the double barlines to some extent. To explain why, I think my main insight is that there is visual redundancy for _short_ barlines (not long ones) with the normal glue, in the vast majority of cases where those short barlines wouldn't be colliding with anything (a rare exception would be a collision with an ison marking). To illustrate what I mean, consider this example:

image

To me this is too disruptive. Consider if that barline were not present:

image

In this case the spacing looks good to me, and I think we could print the short double barline without disturbing the good spacing. We are already taking space for the glue, and since the double barline is a short one, it could easily fit there, without having to reserve extra space for the short double barline. It can just fold into the existing glue (because it is short, and wouldn't collide with anything).

Here is another example to illustrate my point. Consider the spacing here, which looks good to me:

image

Now, when we change the short barline to a short middle barilne, the spacing looks off to me:

image

My intuition here is that this extra space is unnecessary and redundant. If we had left the spacing as it was for the short barline, and just painted an inverted breve on top of it, no collision would have taken place. So reserving the extra space here was redundant, and visually distracting.

basil:
Yet another example in case it helps, the spacing looks good here:

image

But when changing it to a middle barline, redundant extra space is added:

image

Conceptually, the inverted breve could have just been painted above without disturbing the spacing here.

ilizol:
https://github.com/neanes/neanes/pull/2209#issuecomment-4637070694
> I also implemented two spacing/anti-collision mechanisms in the LayoutService. These potentially could have been encoded in the font, but the implementation is significantly simpler and more maintainable in the layout layer.

@danielgarthur
There is a discussion about anticollision.
This is a very impotant feature.
Thanks for your work!

Anticollision Algorithm #262

danielgarthur:
@basil I think this is ready for another round of visual checking.

Before you start your inspection, as a friendly reminder, some of your scores use `spaceAfter` to prevent collisions between barline and gorgons/digorgons/trigorgons. This should no longer be necessary and your scores should look better if `spaceAfter` is set back to `0` in those cases (and probably in all cases)

basil:
Wow, huge improvement. This is a decisive win on the spacing issues around barlines. This is really pushing the envelope of what's possible for Byzantine music typesetting, especially with the formidable challenge of English conventions. Even if we were to never go further than this, I would still be happy with how far we reached.

The remaining issues that I see are minor, and revolve mainly around symmetry around matryriae in the middle of a line (two different issues: root sign imbalance, and lack of lyric tucking), as well as a likely bug when a long lyric has a barline at the end of a line.

---

The last lyric here appears to be off-center:

image

Same with this one:

image

---

Somehow the apostrophe at the top is counted in this martyria's bounding box, but not the root sign at the bottom. As a result it looks off center:

image

---

Should lyrics be allowed to tuck under a martyria? If they are allowed to tuck under the preceding non-martyria neume, I don't see why not. That would make this look less off-balance:

image

---

This one combines the previous two problems:

image

---

Here you can see that the lack of root sign in bounding box pushes the martyria off the page:

image

I could go either way on this: this has a "microtype" hanging punctuation effect that isn't bad at all, but the lack of symmetry from the previous example is what catches my eye (in a negative way). There is no such symmetry issue at the end of a line.

---

The extra space after this gorgon is unfortunate, especially at very low adjustment ratios:

image

Another example at high adjustment ratio:

image

Not sure there is a practical solution other than kerning for that, though. On that note, my eye does occasionally notice the lack of kerning, particularly on lines with very low adjustment ratios (multiple consecutive isons very close together, contrasted with a vareia and apostrophos that are a too far apart), but it is extremely minor and definitely a non-blocking issue. Perhaps that is the logical Phase 2 of this work. I can probably try to help with that at some point. All the font designers I've ever spoken to have stated that kerning is one of the hardest parts of making a font, and something that truly separates a "pro" font from a free font. One even told me that most of what you're paying for in a "pro" font is kerning tables (and, of course, Unicode glyph coverage).

danielgarthur:
@basil I could be wrong, but I think the martyria centering issue is actually that the martyria has glue before and after it, but that glue does not stretch or shrink evenly. My initial thought would be to fix this in the phase 2 loop and center the martyria between the neumes. But you are more knowledgable about the Knuth Plass algorithm, so if this is a terrible idea, let me know.

For the martyria-lyric tuck, in older publications the martyria are much lower so that tucking lyrics under them would be impossible. But should we allow it if the user configures the vertical alignment of the martyria and lyrics such that they fit? Maybe. But I think I will hold off on that for this particular PR.

The lyric offset for transferred bar lines appears to be a pre-existing bug, and I will fix it in this PR.

basil:
The tucking idea isn't essential, it was just one idea I had to solve the symmetry problem. The main visual problem is the lack of symmetry, so if that is solved "somehow" then the specific mechanism by which it is solved is less of a concern.

As far as Knuth Plass goes, conceptually it exerts the same "force" on each item in the line, but the items can stretch differently based on their original widths and stretchability. So to get two items to stretch in parallel, they need to have the same amount stretches (and possibly widths? I'd have to try a few experiments to be sure). That's why I suspected the bounding box was off center - because any small differences in stretches/widths between the two glues would be exacerbated as "force" gets pumped into each element to reach a +1 or -1 adjustment ratio.

So the workaround you proposed of centering them after sounds like it should work, but there might be a way to get Knuth-Plass to center them for you, as long as both glues are symmetrical in their definitions. For non-symmetrical glues to result in the same final widths, I don't believe that is possible with stock Knuth-Plass, although it might be possible with the "continuum width" extension that I've been starting to think about.

danielgarthur:
@basil You were correct. The width/shrink/stretch was not even on either side of the martyria for the `note - martyria - note` case. I am working on resolving that now.

danielgarthur:
@basil I believe the martyria centering is fixed in the case of `note - martyria - note` with no intervening barlines.

The martyria is now centered between the nearest visible content on each side. If lyrics overhang toward the martyria, the lyric edge is used, otherwise the neume ink boundary is used.

I had to use a ruler program to confirm some of the spacing because the eye is deceitful. For example, this is centered, although to my eye it does not look so. But the ruler doesn't lie (I hope).

image

For the overhang on the right margins, that to me seems expected. But I could go either way. I like that the overhang saves space.

image

This is fixed, too.

image

So I think it's all ready for another visual review.

danielgarthur:
@basil I looked into this case.

image

What's happening is that the bounding box is a simple rectangle and the ison's upward curve collides with the naive box. Then may layout changes are adding extra spacing to compensate for this phantom collision.

You can see with the oligon the collision is avoided and the spacing is as defined by the font.

image

The good news is that I've already introduced `collisionRegions` in this PR, but they are only being used for the theseos bar line. But it is easy to expand this concept. I will need to specify multiple collision regions that follow the trigorgon shape more precisely.

So hold off on your visual inspection. I will try to get this done tomorrow.

basil:
OK, I will hold off for now. I did take a quick look tonight and I noticed that martyriae seemed to stop shrinking. I asked Claude Fable 5 Max about this and got this reply:

> The recent "Fix martyria centering logic" commit added a balancing pass that centers a martyria between its neighboring notes by giving the glue on each side equal visible whitespace. But it silently destroyed the boundary's ability to compress during line justification: in every balanced case, both glues ended up with a shrink of exactly zero. Before that commit, each side of a note-martyria-note boundary could give up 0.2 em under compression. The font's engraving defaults are deliberately tuned so that `martyriaGlue.shrink` (0.2) = `martyriaGlue.width` (0.33) - `standardGlue.width` (0.13), meaning a fully compressed martyria gap is supposed to tighten all the way down to ordinary note spacing. After the commit, lines containing a balanced martyria lost that entire 0.4 em budget and had to find all their shrink in the much smaller ordinary inter-note glues, forcing looser line breaks around martyriae.
>
> The shared shrink was computed as `min(fontShrink, leftSlack, rightSlack),` where each slack is the glue's width minus that side's "minimum." The structure of that expression shows the author intended nonzero shrink, but two details made it degenerate. First, the balanced whitespace is defined as the maximum of the two sides' needs, which places whichever side needed more whitespace exactly at its minimum, so one of the two slack terms is always exactly zero, and the `min()` always returns zero. Second, and more fundamentally, those "minimums" were the wrong floors to shrink against: they embed the natural martyria glue width as a structural component, treating a stylistic preference (how much air a martyria gets at natural width) as if it were a hard constraint, when the only genuinely hard constraints are ink-collision clearance, lyric clearance, and measure-bar clearance.
>
> The solution is to keep the shared shrink (both glues compressing in lockstep is what preserves the visual balance at every adjustment ratio) but floor it at the hard minimums instead of the balanced widths: ink and lyric clearance on the left, ink and measure-bar clearance on the right, capped by the font's martyria shrink. With plain neumes this evaluates to the full 0.2 em per side, exactly restoring the pre-balancing compression budget; it only drops below that when a lyric or ink overhang genuinely binds, which is precisely when the old code couldn't compress either. The result is that the balance now holds in both directions (equal stretch and equal shrink on both sides): natural martyria width remains compressible down toward standard spacing as the font intends, and collision floors remain inviolate.

danielgarthur:
@basil Alright, martyria shrink should be restored. I also added collision regions for gorgons and all quantitative neumes. For example:

image

image

This fixes the excessive space after digorgon/trigorgon:
image

basil:
This is very cool. With that last change, do we still need to hardcode extra space after an yporroe with gorgon combinations, or can that be simplified now, since the collision algorithm will deal with any issues stemming from its gorgon combinations?

My point is that these now seem inconsistent:

image

image

image

I think you had shortened the first one earlier as part of a quick and dirty fix. Maybe that isn't needed anymore? Or conversely, should we regard adding whitespace after them to be the "hack" instead, which is now fixed by anti-collision, and therefore remove the extra space from the second and third examples above, to make them more like the first, which is now no longer a "hack" but the correct solution (when combined with anticollision)?

danielgarthur:
@basil Removing the space would send us back to 
image

which won't fix https://github.com/neanes/neanes/issues/2183.

For digorgon and trigorgon, I perused Papa Ephraim's Vespers and Orthros books looking for examples of the spacing used and that is what I came up with.

I'm not saying it has to be that way. But that is why I have it the way it is now. If we want to have space between the apostrofos, but not between oligon and allow oligon to tuck under the gorgon, that's getting into the kerning-pair territory, I think.

Let me know what you think.

basil:
My current opinion is that Papa Ephraim's code from #2183 was a hack, primarily to give his macros better defaults in the absence of an anticollision algorithm, and that the anticollision algorithm now removes the need for that hack. That is consistent with everything Papa Ephraim has ever said to me about his macros (one time, when I told him I had read his collection of macros, he said "you must have had a heart attack to see how messy and awful it was") as well as how I see these scenarios typeset in older publications. So I don't see #2183 as an end unto itself.

From https://analogion.com/psaltologion/neoklis/Tameion_Xourmouziou_A.pdf:

image

image

image

image

From https://stavropoleos.ro/wp-content/uploads/biblioteca-digitala/Psaltichie/psaltichie_carte_veche/pann_heruvicochinonicar_vol1_buc_1847_c5.pdf:

image

image

image

So everything I can see in these very well-typeset publications points away from Papa Ephraim's approach and in favor of a more kerning style approach.

> I'm not saying it has to be that way. But that is why I have it the way it is now. If we want to have space between the apostrofos, but not between oligon and allow oligon to tuck under the gorgon, that's getting into the kerning-pair territory, I think.

Can it just be treated as a regular yporoe, without the gorgon combinations (that is, normal glue before and after it), with the gorgon combinations not affecting spacing in the normal case, but only adding space if there is a collision with the next note, sort of like how we dealt with the mid-barline with the inverted breve.

Ultimately it is your call and this is not blocking feedback, it already looks great and a huge improvement from before. This has gone through a lot of twists and turns and I've learned a lot in the process. If the above is reasonable to implement in the current framework then I think it would be worth it, but if it turns into a massive effort then I would be fine as-is.

danielgarthur:
@basil  I am in favor of ditching the spacing and I've just committed the change to do so.

basil:
@danielgarthur Thank you so much for the enormous amount of effort and care you put into this. I checked my scores, and the problems I pointed out above are fixed. The results look stunning. This is a huge achievement.

src/components/ModeKey.vue
```diff
@@ -118,7 +120,42 @@ const ambitusStyle = computed(() => {
...
+    .map((neume) => NeumeMappingService.getMapping(neume).text)
+    .join('');
+  const font = `${props.pageSetup.neumeDefaultFontSize}px ${props.pageSetup.neumeDefaultFontFamily}`;
+
+  const bounds = TextMeasurementService.getInkBounds(text, font);
+
+  const style = {
+    ...ambitusStyle,
+    marginLeft: `${4 - bounds.inkLeft}px`,
+  } as CSSProperties;
```
basil, line 140: AI says `ambitusStyle` is a `ComputedRef<CSSProperties>`. Vue auto-unwraps refs in templates, but not inside script object spreads. So `...ambitusStyle` spreads the computed ref wrapper, not the CSS object returned by the computed getter. The intended base styles from `src/components/ModeKey.vue`, like `color`, `webkitTextStrokeWidth`, `position`, and `top`, are not included in `ambitusStyleLow` / `ambitusStyleHigh`.

The fix is:

```ts
  const style = {
    ...ambitusStyle.value,
    marginLeft: `${4 - bounds.inkLeft}px`,
  } as CSSProperties;
```

and similarly for `ambitusStyleHigh`.

danielgarthur: Fixed in 37fb67f

basil:
I threw a bunch of agents at this for an hour or so at max settings, and about a million tokens later here is the AI feedback (which may or may not be legitimate).

- The Page Setup preview uses plain object literals and casts them to `NoteElement`, but that cast does not apply the default fields that real `NoteElement` instances have. One preview note includes `measureBarLeft` and `measureBarRight`, so `NeumeBoxSyllable` renders measure bars for it. On this branch, `NeumeBoxSyllable` now also reads computed measure-bar spacing fields like `computedMeasureBarLeftOffsetX` and passes them to `withZoom()`. Because the preview literal does not define those fields, `withZoom(undefined)` emits invalid CSS such as `calc(undefinedpx * var(--zoom, 1))`. The underlying plain-literal preview data was pre-existing on `origin/master`, but the invalid CSS is a regression from the new computed-spacing reads. The best fix is to make the preview use real `NoteElement` instances, or defensively default the computed spacing values to `0` before passing them to `withZoom()`. Using real instances is cleaner because it preserves the component's expected data contract instead of scattering fallback defaults through rendering code.

- `TextBoxRich` can measure rich text before the `NeanesLegacy` font is loaded. The cause is that `TextBoxRich` now renders CKEditor neumes through `pageSetup.neumeDefaultFontFamily + 'Legacy'`, but `TheEditor.vue` only preloads the non-legacy neume fonts. That breaks the existing invariant that all layout-affecting fonts must be ready before documents load. The result can be incorrect rich text height measurement and saved layout churn. The solution is to add `NeanesLegacy`, `NeanesStathisSeriesLegacy`, and `NeanesRTLLegacy` to the editor font preload list.

- Negative neume spacing currently leaks into measure-bar clearance calculations. The problem is that `getTerminalMeasureBarSpacing()` returns raw `getInlineSpacing(pageSetup)`, and `getInlineSpacing()` intentionally allows negative values so users can overlap ordinary neumes. That intentional overlap behavior should not automatically apply to barline clearance, because a terminal or transferred measure bar still needs non-negative room between the bar and adjacent ink. The cause is that the same helper is reused both for ordinary inter-note spacing and for clearance-like barline reservations in penalty widths, terminal finishing glue, and left-bar leading spacing. When `neumeDefaultSpacing` is negative enough, those reservations can shrink below zero or pull a transferred left bar into the following note instead of reserving clearance. The solution is not to clamp `getInlineSpacing()` globally, but to add a separate barline-clearance helper that floors the value at zero and to use it in `getTerminalMeasureBarSpacing()` and related measure-bar clearance paths.

- A martyria's right barline can receive terminal trailing spacing at line end even though nothing reserves that width. The cause is that the martyria breakpoint is encoded with penalty width `0` (`LayoutService.ts:1170`), and `endParagraph`'s `getTrailingNoteReservations()` returns `null` when the last box is a martyria. Later, `centerMeasureBars()` hits its terminal branch for a line-ending martyria with `measureBarRight` and sets `computedMeasureBarRightTrailingSpacing = inlineSpacing` around `LayoutService.ts:4750`. The renderer then applies that computed spacing as inline margin, so the barline is visually moved without a matching line-breaking reservation. On a justified line that breaks after a martyria with a right barline, this free breakpoint can make the barline protrude past the right margin by one inline space, about `3.5px` at `20pt`. The solution is to reserve terminal right-barline clearance for martyria breakpoints the same way note breakpoints do through `getBreakPenaltyWidth()`, or to skip the terminal spacing unless a matching reservation was made.

- The martyria bar-transfer path has a real Phase 1 / Phase 2 width mismatch. The problem is that Phase 1 reserves `barWidth + getTerminalMeasureBarSpacing(pageSetup)` for a bar transferred from a martyria to the first note on the next line (`LayoutService.ts:1150-1156`), but that spacer only includes the base terminal spacing. Phase 2 then shifts the first note left by `barWidth + getMeasureBarLeftLeadingSpacing(...)` (`LayoutService.ts:1611-1621`), which can additionally include an ink-collision deficit. When the transferred-to note has left-overhanging ink at barline height, the Phase 2 shift can exceed the Phase 1 reservation. The visible result is that the transferred barline or its owning note can poke into the left margin by exactly that deficit instead of occupying reserved leading space. The solution is to make the spacer and shift share one transfer-width calculation, either by reserving the collision-aware spacing in Phase 1 or by having Phase 2 use the same base spacing that Phase 1 reserved.

- A line-start martyria ink-overhang shift can break justification by the shift amount. The cause is that Phase 2 adds the martyria's left ink overhang to `currentLine.indentation` for explicit breaks and paragraph starts (`LayoutService.ts:1485-1495`), but Phase 1 computed the line width without that extra indentation. That means the line breaker justifies the contents against one available width, then the renderer effectively moves the whole line right by the overhang afterward. The visible result is limited in scope but real: the code trades a left-margin violation for a right-margin overflow of the same magnitude. This only affects line-start martyrias reached through explicit breaks or paragraph starts, and the size is the martyria's left ink overhang. The solution is to include the overhang adjustment in Phase 1's line-width model, or to apply a compensating width reduction before justification so the post-layout indentation shift is not invisible to line breaking.

- Martyria lyric bookkeeping can use the leading glue width where the trailing pad is needed. The cause is that the martyria path computes separate `useStandardLeadingGlue` and `useStandardTrailingGlue` decisions, but passes `leadingGlueWidth` into `addLyricReservation()` at `LayoutService.ts:1129-1134`. That helper uses the supplied `glueWidth` not only for the pre-element lyric collision check, but also to set `workspace.lyricsEndPx = workspace.neumesEndPx + elementWidthPx + glueWidth`. When `useStandardTrailingGlue !== useStandardLeadingGlue`, the post-martyria lyric end is therefore recorded with the wrong side's glue width. The visible effect should be cosmetic and limited to lyric spacing after martyrias, where the bookkeeping can over-reserve or under-reserve by the difference between standard glue and martyria glue. The solution is to pass the trailing glue width, or to split `addLyricReservation()` so pre-element lyric clearance and post-element lyric-end padding can use different widths.

- A note with a right barline followed by the editor's `Empty` ghost can get mid-line bar spacing even though the bar renders as terminal. The cause is that Phase 1's non-note spacing path still calls `getMeasureBarMinimumGlueWidth(noteElement, nextElement, ...)` when the next element is `Empty`, and that helper uses `2 * inlineSpacing` for any visible measure-bar boundary (`LayoutService.ts:2255-2277` and `LayoutService.ts:5149-5154`). Phase 2 then treats `nextElement == null || nextElement.elementType === ElementType.Empty` as terminal and applies `getTerminalMeasureBarRightSpacing()` instead (`LayoutService.ts:4751-4762`). The result is editor-only over-spacing at the end of the score: line breaking reserves the full mid-line bar minimum, but rendering only needs the terminal `s0` clearance. This should not affect printed/exported output if the `Empty` element is only the editing ghost, but it can make the live editor layout look looser at the final boundary. The solution is to make Phase 1 treat `Empty` like terminal for measure-bar minimum spacing, or to skip the mid-line measure-bar minimum when the next element is the synthetic end ghost.

- Rich-text neume margins are computed for one font but rendered with another. The cause is that `insertneumeutil.Neanes.ts` now asks `FontService` for `Neanes` advance widths, and `Neanes` points to engraving metadata, while `TextBoxRich.vue` renders the same inserted neumes with `NeanesLegacy`. For glyphs whose advance widths changed, especially standalone fthoras, the persisted negative `margin-right` is wrong for the actual render font. This is related to the annotation and preload bugs, but it is a distinct behavior bug. The solution is to compute CKEditor default kerning from the same font family used to render CKEditor neumes, likely the `*Legacy` metadata if the legacy fallback stays.

- Annotations and rich text boxes render inserted CKEditor neumes with different font families. The cause is that `TextAnnotation.vue` still falls back to `pageSetup.neumeDefaultFontFamily`, while `TextBoxRich.vue` was changed to fall back to the legacy family. Inserted neume spans do not persist their own explicit `font-family`, so they inherit the surrounding CKEditor content font stack. That means identical CKEditor neume content can look different depending on whether it lives in an annotation or a rich text box. The solution is to make `TextAnnotation.vue` use the same legacy fallback policy as `TextBoxRich.vue`, if legacy is the intended CKEditor render font.

- Note layout can drift because measurement, collision detection, and rendering build different glyph sequences for the same note. This matters because the font uses contextual substitutions that depend on neighboring glyphs. `getNoteNeumesForMeasurement()` uses the reduced order quantitative, vocal expression, gorgon, accidental, fthora, while `NeumeBoxSyllable.vue` renders fthoras before accidentals and also includes marks such as stavros, time neumes, and koronis. The collision-glyph list has its own ordering too, including placing a tie before an above-style left bar even though the DOM renders the above bar before the tie. When a mark changes or blocks a substitution, the measured width or collision geometry can differ from what is actually rendered. This is a latent trap for future font rules and already risky for legacy fonts whose substitutions can change advance widths. The fix is to introduce one shared helper that returns the canonical rendered neume/glyph sequence for a note, and use it consistently for measurement, ink bounds, collision, and rendering-adjacent logic.

- Collision-box computation is uncached even though it now sits on the layout hot path. The cause is that `getNoteCollisionGlyphBoxes()` performs substitution resolution, mark-offset lookup, and array assembly each time it is called, while adjacent-pair spacing can call it several times through `getNoteVisualMinimumSpacing()` and `getMeasureBarMinimumGlueWidth()`. Phase 2 then recomputes related boxes again from `centerMeasureBars()` and `getNoteBoundsForMeasureBar()`. Widths and ink bounds already have caches such as `neumeWidthCache` and `noteInkBoundsCache`, but collision boxes do not have an equivalent memoized path. For a large score, layout-on-edit can therefore perform thousands of redundant collision-box builds per keystroke. The solution is to add a per-note collision-box cache keyed like the ink-bounds cache, including any bar-reserve override that changes the returned geometry.

- The unanchored mark collision path can spam warnings and use the wrong collision geometry. The cause is that the new collision logic calls `getMarkOffset()` for every mark on every note, often multiple times per layout pass, and several valid marks have no base-glyph anchor pairing. In the engraving metadata, `stavrosAbove` has no pairing with note-base glyphs, and `barlineSingleAbove` plus `barlineShortSingleAbove` pair with fthora glyphs rather than note bases. Every note carrying one of those marks can therefore emit repeated `console.warn` output during ordinary editing. The fallback `{ x: 0, y: 0 }` also places the mark collision box at the base origin, but an unattached zero-advance mark renders at the current pen position, which is the base advance. The result is that collision detection can miss the real mark above the note's right edge while inventing a phantom collision above the left edge. The solution is to remove or rate-limit the warning and to fall back to the pen position, using the base advance, for marks with no anchor pairing.

- The `ModeKey` ambitus positioning has a real scale mismatch. The problem is that `ModeKey.vue` renders the mode key with `element.computedFontSize` and `withZoom()`, but it measures ambitus glyph ink bounds using `pageSetup.neumeDefaultFontSize` at `ModeKey.vue:131-152`. The cause is that the ambitus margin calculations were tied to the global neume default instead of the computed mode-key font size that `LayoutService.ts:688-690` assigns from the mode-key defaults or the element's custom style. This means a custom mode-key font size can produce margins based on the wrong glyph scale, and zooming can make the raw `4px` offsets and the `10px` ambitus gap drift away from the rendered glyph size. The visible result is that the low and high ambitus neumes can sit too close or too far from the parentheses and neighboring glyphs even though the rest of the mode key scales. The solution is to measure with `element.computedFontSize`, and to express the fixed ambitus offsets and gaps through `withZoom()` or an equivalent font-relative value so measurement, rendering, and zoom all use the same scale.

- The `exitsMelismaIntoCenteredLyric` branch is a real documentation drift in the Knuth-Plass notes, even if the code path itself may be intentional. The problem is that `notes/knuth_plass.md` describes same-line melisma-to-syllable spacing through the general `m_i` formula and the melisma collision correction, but `LayoutService.ts:2292-2329` adds a separate case that sets the base width to zero before collision adjustment. The cause is that centered lyrics entered from a carried melisma need a different starting point: the current cursor is already after `spaceAfter`, so the code aligns the next centered lyric's left edge there instead of beginning from the ordinary inline spacing expression. Because the notes otherwise track the implementation closely, future changes to the line-breaking model could miss this exception and accidentally remove or double-count the intended zero-base behavior. The solution is to update `notes/knuth_plass.md` to describe the zero-base melisma exit case explicitly, including when `exitsMelismaIntoCenteredLyric` is true and how the later collision adjustment still applies. It would also be worth adding a focused layout test for a carried melisma entering a centered non-melisma lyric so the behavior is guarded by code as well as documentation.

- `NeumeMappingService.getTextForGlyphName` is dead code. The problem is that the method is exported as part of `NeumeMappingService.ts`, but a repo-wide search only finds its declaration and this review note. The cause appears to be an older API shape that became unnecessary once callers standardized on `getMapping(neume).text` and reverse mapping by SBMufl glyph name. Leaving it in place creates a small maintenance cost because future readers may assume the glyph-name-to-codepoint map is a supported lookup path that needs to be preserved. It also lacks active tests or use sites, so typechecking will not prove that it still behaves correctly if the underlying mapping data changes. The solution is to remove `getTextForGlyphName` and, if glyph-name-to-text lookup is needed later, add it back with a real caller and a targeted test.

- `LayoutService` still contains development scaffolding that should not ship with the layout change. The cause is that `processPages()` can emit an `avg ratio` debug log when `layoutWorkspace.loggingEnabled` is true (`LayoutService.ts:1703`), and the main Phase 1 switch still has comments such as `// PROCESS TEXTBOX`, `// PROCESS NOTE`, and `// PROCESS MARTYRIA`. The log is guarded, so it is not a default production console issue, but it is still a stray diagnostic path in a hot layout service. The `// PROCESS ...` comments also add noise without explaining behavior, ownership, or invariants. This is lower severity than the spacing bugs, but it is the kind of merge-readiness cleanup that prevents temporary review scaffolding from becoming permanent code texture. The solution is to remove the debug log unless there is a real user-facing logging facility for it, and to delete or replace the process labels with comments that explain non-obvious layout logic.

danielgarthur:
@basil Thanks for spending the tokens on that. There were a lot of good catches in there.

I fixed most of them. I do not plan to fix the following.

- Introduce a separate non-negative barline clearance helper instead of using raw `getInlineSpacing()` for measure-bar reservations.
  - I'm not even convinced that we should allow negative spacing anymore. Maybe we change the lower bound to zero in another PR.
 - Add caching for collision-box computation to avoid repeated work during layout.
   - The other caches exist because of the repeated calls to `TextMetrics`. This computation, on the other hand, does not. It is lighter-weight than the cost of creating the key for the cache. Or at least I suspect that is the case. If we see real performance issues, we can optimize then.
 - Clean up `LayoutService` development scaffolding
    - That is left there on purpose.
- Improve handling of unanchored marks
  - I don't think this issue describes a real use case. We prevent marks from appearing above neumes that don't contain anchors before layout service process the pages (via `NeumeReplacements.ts`)
- Create a shared canonical glyph-sequence helper and use it consistently for measurement, collision detection, and rendering.
  - I addressed only part of this (fixing `getNoteNeumesForMeasurement`. But I decided that having consolidated the code was more confusing than leaving it as is.

danielgarthur:
@basil I'm still on the fence about this:

image

Only because overhang is not allowed on the left margin for martyria, so it seems inconsistent to allow right overhang, even though it saves space. Could you please consult any publications that you consider high-quality and see if there is a consistent convention that is followed?

basil:
https://analogion.com/site/pdf/kokkinogorgousa-anthologia-1846.pdf (one of the finest specimens of Byzantine music typography) has them on the inside:

image

image

image

So does https://analogion.com/psaltologion/neoklis/Tameion_Xourmouziou_A.pdf:

image

So the convention appears to be to keep it within the bounding box.

My opinion is that "microtype" style effects like this are cool and modern, and I might even like or use some of them, but if such a feature is developed, it should be applied consistently (both left and right sides), and other things should jut out as well (e.g., hyphens or punctuation of lyrics, etc), and it should be behind a special checkbox in the UI (what InDesign calls "Optical Margin Alignment" or what LaTeX calls "microtype"). The default should be for everything to stay within the bounding box.

basil:
Another round of agent reviews... please feel free to ignore if these aren't legitimate. As you've no doubt observed, many of these agent review cycles devolve into quixotic exercises with diminishing returns. I usually do 2-3 rounds before calling it a day. I also find that Claude is a bit better at this than Codex (but takes a lot more time).

---

- Note-to-note minimum spacing can still be compressed below the collision and barline floors during justification. This is real because each note boundary encodes the same-line floor in the fixed post-break glue `m_i`, but the preceding zero-width pre-break glue still carries ordinary shrink from `standardGlue`. When a line is tight, `positionItems()` can spend that shrink and reduce the actual same-line boundary below the minimum that `calculateInterNoteSpacing()` just computed. This contradicts the documentation claim that the remaining minimum is preserved during justification, and it can reduce a barline or visual-collision floor by up to roughly the standard glue shrink budget. The cause is that shrink is attached to a separate glue item that is not capped against the boundary-specific minimum. Fix this by capping or zeroing pre-break shrink when the following fixed glue represents a hard collision/barline floor, or by moving to an encoding where shrink cannot eat into `m_i`.
- Leading note-to-martyria spacing does not enforce the same measure-bar minimum as martyria trailing and note-to-note spacing. This is real because the martyria trailing path computes `rightSameLineMinimum` with `getMeasureBarMinimumGlueWidth()`, but the leading replacement path only uses `getVisualGlueSpacing()`, lyric reservations, and preserved note reservations. A previous note's right barline is not included in `getElementRightInkOverhang()`, so an ordinary note-to-martyria boundary can shrink toward one inline spacing even when the barline minimum would require more clearance. This makes the behavior asymmetric around martyriae and can show up under tight justification, especially when a right barline precedes a non-right-aligned martyria. The cause is that the leading martyria glue replacement removed the note's trailing glue but did not reapply the structural measure-bar floor for the new boundary. Fix this by incorporating `getMeasureBarMinimumGlueWidth(previousElement, martyriaElement, ...)` into the leading hard minimum and capping `createMartyriaLeadingGlue()` shrink against that value.
- Manual contextual-substitution matching does not honor the font's mark-skipping lookup flag, so collision boxes can disagree with browser-shaped glyphs when marks intervene. This is real because the OTF contextual GSUB lookup uses `LookupFlag 0x100` with `MarkAttachmentType 1`, which lets the browser shaper skip ignored marks while matching contextual rules. `FontService.resolveContextualSubstitutions()` instead walks the glyph-name array by strict indices and requires each backtrack, input, and lookahead class to occupy adjacent array positions. For a sequence such as `yporroi`, an intervening ignored mark, then `gorgonAbove`, the browser can still match the `yporroi -> yporroi.gorgon` lookahead rule, while the manual resolver sees the intervening mark in the lookahead slot and leaves `yporroi` unresolved. `LayoutService.getNoteCollisionGlyphBoxes()` uses that manual result for collision geometry, so note-spacing and barline-collision decisions can be based on a different contextual glyph than the one rendered by the browser. Fix this by exporting lookup flags and mark attachment classes into the metadata and teaching `FontService` to skip ignored marks during contextual matching, or by deriving collision glyphs from the same shaping path used for measurement and rendering.
- Negative note-to-note spacing no longer produces visible overlap, despite the current design note saying it should. This is real for ordinary lyricless note-to-note boundaries because `calculateInterNoteSpacing()` returns `Math.max(baseWidth, minimumWidth)`, and both `getNoteVisualMinimumSpacing()` and `getMeasureBarMinimumGlueWidth()` clamp their minimums at zero. When the computed inline spacing is negative and there is no lyric projection to create a negative tuck, the boundary is floored at zero, so adjacent note boxes touch instead of overlapping. Non-note boundaries can still keep a negative base width in some cases, which makes the behavior inconsistent with the broad documentation statement. The cause is the newer collision/barline floor being applied to all note-to-note boundaries without a policy exception for user-requested negative spacing. Fix this either by preserving negative `m_i` where no real collision/barline floor binds, or by updating the documentation/UI/release notes to state that negative spacing is no longer guaranteed to overlap notes.
- Custom `spaceAfter` and some inline barline cases can add conservative extra spacing because collision helpers measure from the neume run instead of the full note box. This is real because note boxes are added with width `noteElement.spaceAfter + noteElement.neumeWidth`, but `getMinimumSpacingForNoteGlyphBoxes()` is called with `left.neumeWidth` as the left advance. Any positive `spaceAfter` on the left note is therefore not subtracted from the collision equation and can be added again as extra required spacing. A related conservative case exists in the ink-overhang helpers: note overhangs are measured from the glyph run, while inline right barlines are part of the rendered note box, so martyria-boundary spacing can overstate right overhang. These are not collision bugs in the dangerous direction, but they can make manually spaced passages looser than requested. Fix this by passing the actual left box advance into collision spacing and by measuring overhang relative to the same box width used in layout.
- The Page Setup preview does not reflect the user's additive neume spacing adjustment. This is real because `getPreviewGapStyle()` uses `fontSize * glue.width` and omits `form.value.neumeDefaultSpacing`, while the real layout uses `fontSize * standardGlue.width + neumeDefaultSpacing` for ordinary inline spacing. A user changing the neume spacing control can therefore see a preview gap that does not match the score layout. The preview also uses martyria glue based only on element type, so the gap around preview martyriae does not exercise all of the real standard-vs-martyria boundary cases. This is preview-only and does not affect exported or rendered score geometry, so it is lowest priority. Fix it by routing preview gap calculation through the same inline-spacing helper semantics, or by duplicating the additive adjustment in the preview calculation.
- Backtrack glyph order could become wrong if future contextual-substitution metadata contains multi-glyph backtracks. This is latent rather than an active bug because the current metadata has no backtrack sequences longer than one glyph, so the resolver cannot fail on this today. OpenType chain-context backtrack coverage is stored in reverse logical order, but `FontService.contextualSubstitutionMatches()` passes `rule.backtrackGlyphs` directly to `glyphClassesMatch()` and compares it forward against the glyph array. If a future font rule adds a two-glyph backtrack and the metadata preserves the OpenType storage order, the manual resolver would silently miss rules that the browser shaper applies. The cause is that the metadata format does not record whether backtrack classes have already been normalized into forward text order. Fix this by normalizing backtrack class order at metadata generation time and documenting that invariant, or by reversing backtrack classes inside `FontService` if the metadata intentionally mirrors OpenType order.

danielgarthur:
@basil Truthfully, I'm becoming fatigued at this point. I say we sit on this for a few days. I'll look through more of your scores, and you can peruse if you wish, too. If there are no obvious major issues, I say we commit and improve it later.

The only obvious issue that I'm aware of is that bar lines near a martyria that has a wide fthora can collide. I don't know how much it would come in to actual play though. I haven't seen any cases of it happening in your scores, so it may not be common.

But ideally, martyria to bar line collision detection would take into account the fthora, which it currently does not. I started down a rabbit hole of fixing this, but ended up with more bugs than I started with. Maybe I'll try it by hand if it's really a major issue.

basil (dismissed):
Agreed, this is as good a place as any to stop. Great job!

danielgarthur:
@basil Okay. I couldn't let it rest. I fixed bar line spacing and collision-handling on the note-martyria and martyria-note boundaries in the case where the note owns the bar line.

image
image

I think this should be good enough since if someone just so happens to put a non-short bar line on a martyria that has a wide fthora on it and it collides... well they can just put the bar line on the note instead. Long term, I think the bar lines need to be somehow not owned by the note or martyria but become first class citizens. I think you mentioned something similar at one time. But that is another task for another time.

Hopefully there was no collateral damage... I'll check scores again.

danielgarthur:
@basil You may want to look over knuth_plass.md changes and have an agent update it more accurately/tastefully.

basil:
I've deployed this change and rebuilt all my scores on my web site. Everything looks stunning! It's difficult to express how grateful I am. This was a huge achievement. Fantastic work, and thank you!

PR #2215, opened 2026-06-04
feat: Port UI to `shadcn-vue`

### LINKED ISSUE #2208: Component framework
The Neanes UI is built almost entirely from hand-rolled components. This approach is simple and lightweight, but it has some drawbacks:

- The UI has an inconsistent look, since each component is built independently.
- New UI work is painful, because every needed component must be built from scratch.
- The code is hard to maintain, with styling copied and pasted across many places.
- Features are missing, like keyboard accessibility and tab context menus in the web application.

It would be desirable to improve or replace the current components with off-the-shelf ones that provide the features we need in a consistent design language without introducing bloat or maintenance burdens. Things we need:

- Buttons and button groups (e.g., Auto/Insert/Single on the main toolbar)
- Toggles and toggle groups (e.g., for text box settings)
- Toolbars
- Buttons integrated with a generic icon library such as [Lucide](https://lucide.dev/) (for the toolbar)
- Modal dialog boxes
- Form fields: checkboxes, selects, and inputs with labels (e.g., for the preferences dialog)
- Number fields and sliders
- A full menu bar for the web application
- A right-click tab context menu for the web application
- Tooltips with keyboard shortcuts (for hovering over toolbar items)
- Separators
- Color picker

The current best-of-breed library for this in Vue appears to be [Reka UI](https://reka-ui.com/). It provides most of these components in unstyled "headless" form, with standard, well-tested implementations of keyboard navigation, accessibility, and focus management. This provides a great set of building blocks for building your own styled UI from scratch or for building an opinionated set of styled components. I think we could take two approaches:

1. We could refactor the existing application into a set of shared components based on Reka, consolidating the various copy-pasted styles and logic. This is possible but a lot of work.
2. An easier approach, and one which requires less design skill, is to adopt some preëxisting set of styled components built on top of Reka, then convert the application one toolbar or dialog at a time. Examples include [Nuxt UI](https://ui.nuxt.com/), [`shadcn-vue`](https://www.shadcn-vue.com/) (very popular in the AI vibe coder community), [UI Thing](https://ui-thing.behonbaker.com/), and [Typlog UI](https://ui.typlog.com/).

## Open questions

Which approach is most maintainable and most in line with the project's aesthetics and developer preferences? I'm out of my depth here and not really qualified to have a strong opinion.

`shadcn-vue` pulls in the preferred "vibe coder's stack" of 2026 (Tailwind CSS plus copy-pasted components) but little else in the way of dependencies. This might actually be a good fit for us, since the dependencies are light and the components can be easily modified for our needs. However, I can imagine the copy-paste model and the use of Tailwind becoming a maintenance burden. Libraries like Nuxt UI feel more formalized and maintainable but are also much heavier in terms of dependencies.

I'd like to hear from others with more experience here. One thing seems clear: it isn't practical to keep improving the UI on the current stack, both because every new component must be developed from scratch and because there are no clear, reusable building blocks from which to derive new functionality. Something should be done in the long term, but I'm not sure which direction we should take.

basil:
This PR is a first cut of migrating the UI to https://www.shadcn-vue.com, an implementation of the Reka UI APIs. This is the most lightweight framework I could find for adding a UI to the app. It involves:

- Adding Reka UI and Tailwind as core dependencies
- Use the shadcn-vue CLI to copy and paste the wrapper components into your codebase. This is mildly horrifying, but on the plus side you can pick and choose exactly which components you want, so you don't have to pull in a charts library if you don't use charts. The wrapper components are basically thin calls to Reka that add Tailwind styling.
- Some of the components themselves have additional dependencies, for example the toast component requires `vue-sonner`.

From a maintainability perspective, the critical thing is not to edit or modify the wrapper files, a discipline which I stuck to with this PR but which was not enforced. If we do move forward with shadcn-vue, I'd want to treat the `src/components/ui` like a `vendor` dependencies directory, with some kind of enforcement that it never be modified, except when upgrading `shadcn-vue`. To upgrade, you just run the same CLI again (the one you used to copy and paste the components in) and it just updates them with newer versions, typically just updating Tailwind CSS rules. The dependency surface area is otherwise about as light as it gets for Vue UI frameworks: Tailwind and Reka. I didn't find an easy way of getting rid of Tailwind. It would have been possible to do a hard fork of the `shadcn-vue` components to remove Tailwind, but then that would mean we couldn't get updates to them. Ultimately I judged the displeasure of Tailwind to be worth the benefit of free upgrades to shadcn-vue.

The main benefits of Reka are a tried and proven implementation of basic accessibility, keyboard navigation, focus, etc. The main benefits of shadcn-vue are a decent UI that requires relatively little effort to put together using AI, even for someone like me with absolutely zero UI building experience. As part of this effort, I:

- Redesigned all existing dialog boxes
- Added right-click context menus to the tab bar of the web app
- Added a new help about page
- Updated tooltip display to show prettier keyboard shortcuts
- Improved the display of input units
- Moved zoom errors to use toasts

I attached screenshots below. I think the result looks a lot better, and more importantly it replaced a lot of bespoke internal logic with more standardized equivalents. For example, I replaced the "Common Combos" with a standard accordion feature. We now get keyboard navigation in menus in the web app for free (as well as support for submenus there) and a host of other improvements.

I've attached screenshots below of how this is looking so far. If we do want to go down this path, then the next steps after this I think would be to clean up the toolbars, which have become a dumping ground for all sorts of non-toolbar stuff that should probably go in separate popovers or dockable bars. Those should be a lot easier to design and implement with shadn-vue, as well.

image

image

image

image

image

image

Fixes #2208


basil:
Taking this out of draft, as I tested this manually and I think it is working pretty well. Also the agent reviewers couldn't find anything wrong with it either.

basil:
The risk associated with Tailwind is low: If it ever disappears, it would be fairly easy to do a mechanical conversion into plain CSS. Similarly, the risk of shadcn-vue itself disappearing is low for us because its model involves copying and pasting the shadcn-vue components (really just thin wrappers around Reka + Tailwind styling) into the repository itself, so upstream disappearing wouldn't affect us.

The main risk associated with this PR is Reka itself disappearing, since that is where most of the complex logic lives, and it would be nontrivial to port shadcn over to a different framework. Mitigating this risk is the fact that Reka is a low-level library used by 3 separate UI frameworks (shadcn-vue, Nuxt UI, and UI Thing), one of which is backed by Vercel, so it it has a higher chance of surviving than any individual UI library. I am sure that it, too, will fade eventually, but the hope is that due to the clean separation of concerns between low-level API and styling, as well as a lightweight set of dependencies, that it will be easier to migrate to whatever comes next compared to a heavyweight all-in-one megaframework.

danielgarthur:
Another consideration, which I am quickly realizing, is that by the time the library stops being maintained, AI will be even more capable of migrating to the next UI framework...

basil:
@danielgarthur Can I ask you to test this and let me know if you find any regressions? I've already tested it all I can I think.

basil:
In some recent commits, I switched the shadcn-vue icon theme to [Phosphor](https://phosphoricons.com/) (more compatible with the skeumorphic design of this app) and sprinkled icons throughout all the dialog boxes. I think it makes the app feel a lot more pleasant, and having the entire icon set built in should be a plus for new feature development.

danielgarthur:
Overall it looks good. I agree that copying over all the components is horrifying. Do we need to copy all those components? Are we ever going to use calendars, for example?

I could only find one issue, and it's not exactly a regression.

- When setting intervals for playback, if the intervals don't sum to 30, only the first offending interval row is highlighted, even if multiple rows are in the error state. This was wrong before, but it's more obvious now because of the red highlighting.

danielgarthur (dismissed):
Only concern is whether we need _all_ the components.

basil:
Good point. I deleted the calendar and range calendar components, as well as the pin input component. I tried to delete the sheets component as well, but it was used as a transitive dependency and came back when I ran `./scripts/update-shadcn-vue.sh`. I kept a few other unused components as they seemed like we could use them in the future.

PR #2249, opened 2026-06-10
feat: Dockable workspace

### LINKED ISSUE #2092: Lyric box adjustable height
**Is your feature request related to a problem? Please describe.**
Writing lyrics for pieces that are longer than 2-3 pages because quite annoying when the lyric text book so so small and not adjustable.

**Describe the solution you'd like**
Make the lyric box (ctrl+L) height adjustable.

**Describe alternatives you've considered**
Grabbing the lyrics into a notepad to be able to adjust lyrics quicker, however you lose the benefit of seeing the music adjust to the lyrics in real time, and lose efficiency due to constantly copying and pasting back and forth.

Thank you @danielgarthur your software is absolutely legendary.

### LINKED ISSUE #2217: Clean up toolbars
The bottom toolbars have become a dumping ground for generic properties that don't belong in toolbars: various checkboxes, input fields, selects, etc. These belong in a vertical view, not a horizontal one, where they can be styled properly. A natural place to put these would be a new properties panel, like the one in MuseScore, which contains the properties for the currently selected object: https://handbook.musescore.org/basics/properties-panel

With properties moved to a dedicated properties panel, the toolbars at the bottom of the screen could be shrunk down to the actual toolbar content:

- Neume toolbar: two rows of buttons
- Martyria toolbar: one row of buttons
- Drop cap/text box toolbars: Font selection, bold/italic, font size, etc (these could also be properties as well, with more advanced options living in the properties)

The natural home for the properties panel would be on either the left or right side of the screen. Rather than hard-code yet another layout option, my proposal would be to introduce the notion of vertical panels:

```
+---- Main toolbar -- plain DOM, top, fixed (FileMenuBar + ToolbarMain) ---+
|--------------------------- <DockviewVue> --------------------------------|
| +------------++-------------------------------------++-----------------+ |
| | left edge  ||        center grid                  ||   right edge    | |
| |  NEUME     ||       SCORE TABS                    ||   PROPERTIES    | |
| | SELECTOR   || reorder / close / right-click / new ||    (new panel)  | |
| | (panel)    ||  (single live surface     )         ||                 | |
| +------------++------------------------------++------------------------+ |
|-- Contextual toolbar -- plain DOM, bottom, fixed (swaps on selection) ---|
+--------------------------------------------------------------------------+
```

This would be implemented with [Dockview-Vue](https://dockview.dev/), a zero-dependency library which would replace our dependency on `vue3-tabs-chrome`. Probably the syllable positioning dialog should also become a panel as well. The idea is we would have three panels: neume selector, syllable positioning, and the new properties panel. By default the neume selector panel would remain on the left, as it is today, but it could also be moved to the right, or floated:

Image

The syllable positioning panel would also live on the left as another tab, not displayed until you click on it. The properties panel would live on the right by default and would be open by default. You could float it, or move it to the left if you wanted. If you move both the neume selector and properties panels to the left, you would likely be switching between them often, so it makes the most sense for the panels you use regularly to be either floating or docked to separate sides of the screen.

An extension to this design would be to allow the contextual toolbar to be docked to the top (under the main toolbar), bottom (where it currently lives), or floating, thus providing the user with maximum flexibility for commonly used buttons.

Likely depends on #2208, as designing the new properties panel would be far more difficult without a component framework.

### LINKED ISSUE #2259: Automatic hide left side panel
the ability to hide and show the left panel is crucial when you need more space on a small screen like laptop to type.

basil:
## Summary

This PR replaces the editor's fixed layout (static left panel + stacked bottom toolbars) with a dockable pane workspace built on [dockview-vue](https://dockview.dev/). Editor tools now live in five panes that can be shown/hidden from a new **View** menu, docked as tabs in collapsible left/right drawers, dragged between edges, or floated as windows over the score. Full element settings move out of the bottom toolbars into a new **Properties** pane; the bottom toolbars remain as slim, contextual quick-controls.

## User-facing changes

image

### New View menu

A **View** menu is added to both the native Electron menu and the in-browser menu bar, with checkbox toggles for each pane and a **Reset Layout** item that restores the default arrangement:

| Pane | Default | Home edge | Notes |
| --- | --- | --- | --- |
| Neume Selector | visible | left | Same panel as before, now dockable |
| Common Combinations | hidden | left | Was an accordion inside the neume selector; now its own pane |
| Properties | hidden | right | **New** full type-specific settings for the selected element |
| Selection | hidden | right | **New** common selection settings (see below) |
| Lyrics | hidden | right | Replaces the lyric manager bottom toolbar |

- `Edit -> Lyrics` (<kbd>Ctrl</kbd>+<kbd>L</kbd>) moves to `View -> Lyrics`; the shortcut still works and now toggles the Lyrics pane.
- The previously dev-only "View (Debug)" menu is folded into the new View menu (reload/devtools/zoom items still appear only in development).
- The native menu checkboxes stay in sync with the actual pane state via a new IPC channel; the renderer is the source of truth.

### Pane behavior

- Panes dock as tabs inside collapsible edge drawers on the left or right (top/bottom are intentionally not allowed, and per-pane allowed edges are enforced during drag and drop).
- Panes can be floated three ways: the header button, the tab context menu, or dragging a tab onto the editor center (the pane floats at the drop point). Floating panes auto-size to their content once when floated, capped at 75% of the available height, and can then be resized freely; content scrolls internally.
- Hiding a pane collapses its drawer (or re-docks a floating pane to its last-used edge); each pane remembers the last edge it was docked on.
- Dropping a pane on a root edge re-docks it into the permanent drawer for that edge instead of creating dockview's default grid splits. Whole-group drags, external drag payloads, and unsupported drop zones are rejected, including their drop overlays.
- Pane titles are localized and re-translate live on language change.

### Properties pane

Type-specific settings that previously lived in the bottom toolbars now appear in `View -> Properties`, with one component per element type: neume, martyria, mode key, tempo, text box, rich text box, drop cap, image box, and lyrics. This includes settings such as martyria note/scale overrides and BPM, neume positioning and `Accepts Lyrics`/`Ignore Attractions`, mode key size/color/margins/`Permanent Enharmonic Zo`, text box formatting, image box dimensions, and chromatic fthora note.

### Selection pane

`View -> Selection` shows settings common to any selected element: an element-type badge, **Section Name** (moved from the toolbars), **Line Break** (now a dropdown: None/Left/Center/Justify), **Page Break**, **Copy Element Link**, and **Delete**. For a range selection it shows the element count and a delete button. It handles alternate-line elements correctly (breaks are disabled there, and deleting the last alternate-line element removes the alternate line itself).

### Slimmed bottom toolbars

The bottom toolbars are now quick-access strips for the most common actions, shown for neumes, martyriae, mode keys, lyrics, text boxes, and drop caps:

- Neume/martyria/mode key toolbars keep their sign buttons (gorgon, fthora, klasma, tempo signs, measure bars, alignment, etc.); their form fields (BPM, offsets, note/scale overrides, section name, ...) move to the Properties/Selection panes.
- Lyrics/text box/drop cap toolbars keep font family, size, and bold/italic; everything else (default-style checkbox, color, outline, line height, width, line span, ...) moves to the Properties pane.
- The tempo, image box, and rich text box toolbars are removed entirely (`ToolbarTempo.vue`, `ToolbarImageBox.vue`, `ToolbarTextBoxRich.vue` deleted). These elements are now configured exclusively through the Properties pane.
- `ToolbarLyricManager.vue` is deleted in favor of the Lyrics pane, which keeps the same features (lock lyrics, Save Current Melismas, auto-syncing textarea).

### `WorkspaceDockLayout.vue` (new)

The core integration. The editor is a permanent, headerless dockview center panel and each tool pane is a dockview panel inside named edge groups:

- **Drop policy** is centralized in a `decideDrop` state machine driving both `onWillShowOverlay` (so rejected targets never show an overlay) and `onWillDrop`. Center-content drops on the editor are translated into "float at pointer"; root-edge drops are translated into "move into the permanent edge group"; everything else either passes through to dockview's native behavior (with drawer expand/collapse scheduled around it) or is rejected.
- The editor group's content drop zones are constrained to a single center zone via a typed reach-in to dockview internals (no public per-group API exists); this is documented inline.
- Pane visibility is *derived* from dockview state (active tab + drawer collapsed + location) and emitted upward, coalesced per animation frame, so the View menus always reflect reality regardless of how the change happened (menu click, header button, tab context menu, drag, drawer collapse).
- Custom tab, header-actions (float/dock + hide/show), and tab context menu components replace dockview's close-oriented defaults, since these panes are persistent app surfaces.
- Whole-group drags from the blank header area are blocked with capture-phase listeners so only single tabs can be dragged.

### Printing

The center editor must keep dockview's default `onlyWhenVisible` renderer. `renderer: 'always'` silently breaks printing (documented inline). A dedicated `@media print` block flattens the dockview ancestor chain for the center editor and hides the rest. This fixes two real print bugs caused by dockview's window-derived inline pixel geometry: Blink's shrink-to-fit scaling the whole printout to ~67%, and the page flow being clipped to a single page by a fixed-height `overflow: hidden` ancestor. The selector set is audited against dockview-core 6.6.1; the inline comment flags it for re-audit on dockview upgrades.

### `InspectorContext` (new)

A discriminated union (`none` / `neume` / `martyria` / `mode-key` / `tempo` / `text-box` / `rich-text-box` / `drop-cap` / `image-box` / `lyrics` / `range`) computed once in `TheEditor.vue` from the existing selection state, with a `source: 'score' | 'header-footer'` discriminator for text boxes. It drives the Properties pane, the Selection pane, and the contextual bottom toolbar from a single source of truth. The element type-guard helpers were converted to TypeScript type predicates to support this.

### Lyrics assignment is now per-workspace

The old global `throttle(assignLyrics)` is replaced with a per-workspace debounce keyed by workspace id (`pendingLyricsAssignmentTimers`). Pending assignments are flushed on save and on workspace switch, and canceled when a workspace closes. `assignLyrics`/`extractStaffLyrics` now operate on an explicit workspace, so a debounced assignment that fires after switching tabs updates the *correct* (possibly background) workspace (marking it unsaved) instead of leaking into the selected one, and DOM-coupled lyric updates are skipped for non-selected workspaces.

### Other editor changes

- Keyboard handling: pane content is wrapped with `data-editor-shortcuts="ignore"`, so typing in pane inputs no longer triggers editor shortcuts (generalizes the previous toolbar-only special case).
- `pageBackgroundRef`/`tabsRef` access is now null-safe because the center editor mounts inside dockview (e.g. zoom-to-fit, scroll restore, and tab add/remove no longer assume the refs exist).
- `replaceScoreElement` re-points the selection at the replacement element so the Properties/Selection panes don't go stale after replace.
- Dockview mounts pane content through its own renderer, so panes get their own `TooltipProvider`.

### IPC / menu plumbing

- `FileMenuLyrics` is replaced by `FileMenuViewPaneVisibility { paneId, visible? }` and `FileMenuViewResetPaneLayout` (main - renderer).
- New `SetWorkspacePaneVisibility` (renderer - main) keeps native menu checkboxes in sync; menu clicks optimistically revert and wait for the renderer to confirm.

### Models

- New `src/models/WorkspacePane.ts`: the pane registry (ids, home/allowed edges, default sizes, default visibility, i18n title selectors) shared by the renderer and the Electron main process.
- `Workspace.lyricManagerIsOpen` is removed; Lyrics pane visibility is global workspace UI state, not per-document state.

### i18n and docs

- New keys across `en`/`el`/`id`/`ro`: `menu.view.*` (pane names, Reset Layout), `toolbar.properties.*`, `toolbar.selection.*`, `toolbar.workspace.*` (dock/float/hide/show), `toolbar.common.alignment`, `menu.insert.dropCap`. The now-unused `editor` namespace (`neumeComboHeader`) is removed.
- The user guide (`docs/guide/index.md`, `docs/guide/advanced.md`) is updated: a new "Workspace panes" section, and all references to the lyrics toolbar / bottom-toolbar settings now point at the Lyrics, Properties, and Selection panes. A few typos are fixed along the way.

## Testing Done

1. `npm install && npm run dev`.
2. View menu: toggle each pane from `View`, via the pane header `x`/eye buttons, and via the tab context menu -- the menu checkmarks should track all of them. <kbd>Ctrl</kbd>+<kbd>L</kbd> should toggle the Lyrics pane.
3. Drag panes: between the left and right drawers, onto the editor center (should float at the cursor), onto a root edge while the drawer is collapsed (should re-dock), and confirm top/bottom edges are rejected with no overlay. `View -> Reset Layout` should restore the defaults.
4. Select each element type (neume, martyria, mode key, tempo, text box, rich text box, drop cap, image box, lyrics) and verify the Properties pane, Selection pane, and the slim bottom toolbar; verify section name, line/page break, copy element link, and delete from the Selection pane, including inside an alternate line and with a range selection.
5. Lyrics pane: typing assigns lyrics (debounced); switch tabs mid-typing and confirm the edit lands in the originating workspace and marks it unsaved; lock lyrics and Save Current Melismas still behave as documented.
6. Print/export to PDF with panes open, floating, and hidden -- the score should print at full scale, with all pages, and with no pane chrome.
7. Typing in pane inputs must not trigger editor shortcuts; switching the language should re-translate pane titles and menus.

Fixes #2217
Fixes #2092
Fixes #2259


danielgarthur:
Overall this looks good.

Minor nitpick: The position of the element type catches my eye as odd. Not sure where it should be exactly but not there.

image

danielgarthur:
Regression: keyboard shortcuts like Copy/paste/select all/etc. does not work in lyrics manager or other inputs like selection name. This is working on `master`.

basil:
> Overall this looks good.

Thanks!

> Minor nitpick: The position of the element type catches my eye as odd. Not sure where it should be exactly but not there.
> 
> image

Tried shadcn-vue vertical instead of horizontal and it looked worse:

image

Keeping as is, as I think this is the best of the available shadcn styling.

> Regression: keyboard shortcuts like Copy/paste/select all/etc. does not work in lyrics manager or other inputs like selection name. This is working on `master`.

Thanks, fixed in commit 845285a5.

danielgarthur:
What about `margin-left: auto;` on the badge?

image

basil:
Thanks, fixed in commit f05b1c62.

danielgarthur:
I am also seeing an "extra blank page" bug. It is caused by these divs here that appear when I open a right panel like "Selection" (maybe other stuff triggers it too).

image

basil:
Thanks, fixed in commit 2749ca9a

danielgarthur:
I think the ga-vou in the bottom right corner should go on the first line either before or after neume select

image

danielgarthur:
Gap Above and Below inputs should be the same width. This is in Text Box, but may be happening elsewhere.

image

basil:
> I think the ga-vou in the bottom right corner should go on the first line either before or after neume select

Neume select isn't always displayed, though.

basil:
> Gap Above and Below inputs should be the same width. This is in Text Box, but may be happening elsewhere.

Fixed in commit 9a5df99b

basil:
> > I think the ga-vou in the bottom right corner should go on the first line either before or after neume select
> 
> Neume select isn't always displayed, though.

See https://github.com/neanes/neanes/issues/2274 for my thoughts about how to deal with this in the long term.

basil:
> I think the ga-vou in the bottom right corner should go on the first line either before or after neume select

@danielgarthur Is this a blocking issue, and if so you could you please provide the reasoning for why you think the above?

danielgarthur:
No it's not blocking.

basil:
Sounds good. I'm happy to make any additional changes here if there's anything else you’d like changed. I think the more complete solution for the feedback around current note display probably belongs in a separate follow-up to build a dedicated lower status bar, since that needs a bit more design work. Separately, I also have a CKEditor cleanup PR stacked on this one to move it to the standard shadcn-vue UI, so once this PR feels settled I can open that up as well.

danielgarthur:
It's good. I agree that the note names belong somewhere else altogether. But that can be done later.

basil:
Cool, I'll hold off for now until you feel comfortable merging this. Let me know when you feel ready and then I'll merge and open the follow-up PR for CKEditor.

basil:
@danielgarthur I checked in a design doc in commit 1aa7aa6954c4d2cd8e143c1469de0107746719f3. Not sure what your perspective is on checking in slop like this. On the one hand, it is somewhat laborious to keep it maintained when doing a major subsystem refactor/rewrite, as you've no doubt discovered on the tight font branch. On the other, it does help a human to quickly understand the code. While a human could easily regenerate it with AI, it would take a lot of time and tokens, and probably an hour or so of turns to come up with a high-quality doc like the one I checked in, so I think it makes sense to check it in once it's done. Even more valuable than human consumption, though, is (I think) AI consumption, as an AI can read the doc quickly to get up to speed with a given subsystem before diving into implement a feature. I've observed this firsthand as I hand these subsystem-level design docs to AI which can read them instantly and then start implementing new features without confusion. I've also been making sure these docs explicitly document interactions with third-party dependencies, so that when we upgrade these dependencies AI can read the doc to figure out what behaviors we rely on and then validate that the upgrade is safe.

danielgarthur (approved):

danielgarthur:
> Cool, I'll hold off for now until you feel comfortable merging this. Let me know when you feel ready and then I'll merge and open the follow-up PR for CKEditor.

Feel free to merge.

basil:
> It's good. I agree that the note names belong somewhere else altogether. But that can be done later.

⇨ #2301

PR #2253, opened 2026-06-10
Options API migration fixes

### LINKED ISSUE #2241: Mode key tempo no longer uses `setModeKeyTempo`
Found a minor regression from the move to the Options API in #1757. `ToolbarModeKey` emits a generic update for the tempo menu:

```vue
<!-- src/components/ToolbarModeKey.vue -->
<ButtonWithMenu
  :options="tempoMenuOptions"
  :tooltip="$t(($) => $.toolbar.common.tempoSign, { ns: 'toolbar' })"
  @select="$emit('update', { tempo: $event } as Partial<ModeKeyElement>)"
/>
```

`TheEditor` handles that generic update with a plain property update:

```vue
<!-- src/components/TheEditor.vue -->
<ToolbarModeKey
  ...
  @update="updateModeKey(selectedElement as ModeKeyElement, $event)"
  @update:tempo="
    setModeKeyTempo(selectedElement as ModeKeyElement, $event)
  "
/>
```

But `ToolbarModeKey` does not emit `update:tempo` anymore. It still declares it in `defineEmits`, and `TheEditor` still listens for it, but the listener is orphaned.

The orphaned `setModeKeyTempo` path did two things that `updateModeKey({ tempo })` does not:

```ts
function setModeKeyTempo(element: ModeKeyElement, neume: TempoSign) {
  if (element.tempo === neume) {
    updateModeKeyTempo(element, null);
  } else {
    updateModeKeyTempo(element, neume);
  }
}

function updateModeKeyTempo(element: ModeKeyElement, tempo: TempoSign | null) {
  let bpm = element.bpm;

  if (tempo != null) {
    bpm =
      editorPreferences.value.getDefaultTempo(tempo) ??
      TempoElement.getDefaultBpm(tempo);
  }

  updateModeKey(element, { tempo, bpm });
}
```

So the current generic path has two regressions:

- It sets `tempo` but leaves `bpm` unchanged instead of applying the user/default BPM for the selected tempo sign.
- It cannot toggle an existing tempo off by selecting the same sign again.

### LINKED ISSUE #2242: Right align martyria no longer clears `quantitativeNeume`
Found a minor regression from the move to the Options API in https://github.com/neanes/neanes/pull/1757. The martyria right-align button emits a generic update:

```vue
<!-- src/components/ToolbarMartyria.vue -->
@click="
  $emit('update', {
    alignRight: !element.alignRight,
  } as Partial<MartyriaElement>)
"
```

`TheEditor` applies it through the generic martyria updater:

```vue
<!-- src/components/TheEditor.vue -->
<ToolbarMartyria
  ...
  @update="updateMartyria(selectedElement as MartyriaElement, $event)"
/>
```

The selected-martyria keyboard path does the same thing:

```ts
throttled.updateMartyria(martyriaElement, {
  alignRight: !martyriaElement.alignRight,
});
```

The original behavior was:

```ts
updateMartyriaAlignRight(element: MartyriaElement, alignRight: boolean) {
  this.updateMartyria(element, { alignRight, quantitativeNeume: null });
}
```

That behavior was introduced by `39f91cd4` (`Add option for quantitative neume after martyria (#1519)`) when `quantitativeNeume` was added. It intentionally cleared `quantitativeNeume` whenever alignment changed.

Current `origin/master` updates only `alignRight`, so stale `quantitativeNeume` survives alignment toggles. `NeumeBoxMartyria` only renders the quantitative neume when both fields are set:

```ts
const hasQuantitativeNeume = computed(
  () => props.neume.quantitativeNeume != null && props.neume.alignRight,
);
```

So if a martyria has a quantitative neume and the user turns right-align off:

- the quantitative neume disappears visually because `alignRight` is false;
- the saved model still contains `quantitativeNeume`;
- turning right-align back on can resurrect the stale quantitative neume.

### LINKED ISSUE #2243: Updating martyria scale with keyboard shortcut no longer makes the martyria manual
Found a minor regression from the move to the Options API in https://github.com/neanes/neanes/pull/1757. For selected martyria elements, the keyboard config path updates scale through the generic martyria updater:

```ts
// src/components/TheEditor.vue
} else if (martyriaConfigMapping.scale != null) {
  throttled.updateMartyria(martyriaElement, {
    scale: martyriaConfigMapping.scale,
  });
}
```

The relevant mappings are reachable through the martyria scale modifier (`KeyZ`) in `NeumeKeyboard`.

Before the Options API commit, scale changes went through:

```ts
updateMartyriaScale(element: MartyriaElement, scale: Scale) {
  if (element.scale === scale) {
    return;
  }

  this.updateMartyria(element, { scale, auto: false });
}
```

The important side effect was `auto: false`. Changing a selected martyria's scale is a manual override; leaving `auto` true leaves the model internally inconsistent with that user action.

The selected-martyria keyboard branch also contains:

```ts
if (martyriaConfigMapping.note != null) {
  // This case will not currently happen
  // because no keyboard mapping exist for it
  throttled.updateMartyria(martyriaElement, {
    note: martyriaConfigMapping.note,
  });
}
```

If this branch is reached, it has the same lost behavior. The historical handler was:

```ts
updateMartyriaNote(element: MartyriaElement, note: Note) {
  if (element.note === note) {
    return;
  }

  this.updateMartyria(element, { note, auto: false });
}
```

Because the current code itself says the note branch should not happen, the strongest confirmed remaining user-facing bug here is the scale path. The note branch should still be fixed or removed because it is stale and would preserve `auto = true` if invoked.

basil:
Fixes #2241
Fixes #2242
Fixes #2243

- Martyria keyboard note/scale overrides now set `auto: false`, restoring the old `updateMartyriaNote` / `updateMartyriaScale` behavior.
- Martyria alignment changes now clear `quantitativeNeume` whenever `alignRight` actually changes, covering both toolbar and keyboard paths through generic `updateMartyria`.
- Mode-key tempo selection now emits `update:tempo` again, so the parent calls `setModeKeyTempo` and recomputes BPM via `updateModeKeyTempo` instead of only writing tempo.

### DIFF
--- a/src/components/TheEditor.vue
+++ b/src/components/TheEditor.vue
@@ -2284,11 +2284,15 @@ function onKeydownNeume(event: KeyboardEvent) {
           // because no keyboard mapping exist for it
           throttled.updateMartyria(martyriaElement, {
             note: martyriaConfigMapping.note,
+            auto: false,
           });
         } else if (martyriaConfigMapping.scale != null) {
-          throttled.updateMartyria(martyriaElement, {
-            scale: martyriaConfigMapping.scale,
-          });
+          if (martyriaElement.scale !== martyriaConfigMapping.scale) {
+            throttled.updateMartyria(martyriaElement, {
+              scale: martyriaConfigMapping.scale,
+              auto: false,
+            });
+          }
         } else if (martyriaConfigMapping.martyriaAlignmentToggle === true) {
           throttled.updateMartyria(martyriaElement, {
             alignRight: !martyriaElement.alignRight,
@@ -4237,10 +4241,16 @@ function updateMartyria(
   element: MartyriaElement,
   newValues: Partial<MartyriaElement>,
 ) {
+  const values =
+    newValues.alignRight !== undefined &&
+    newValues.alignRight !== element.alignRight
+      ? { ...newValues, quantitativeNeume: null }
+      : newValues;
+
   commandService.value.execute(
     martyriaCommandFactory.create('update-properties', {
       target: element,
-      newValues: newValues,
+      newValues: values,
     }),
   );
 
--- a/src/components/ToolbarModeKey.vue
+++ b/src/components/ToolbarModeKey.vue
@@ -112,7 +112,7 @@
     <ButtonWithMenu
       :options="tempoMenuOptions"
       :tooltip="$t(($) => $.toolbar.common.tempoSign, { ns: 'toolbar' })"
-      @select="$emit('update', { tempo: $event } as Partial<ModeKeyElement>)"
+      @select="$emit('update:tempo', $event)"
     />
     <span class="space" />
 


basil:
@danielgarthur This PR (literally) robotically restores the old behavior prior to #1757, but I'm not sure if the changes listed above were intentional changes in behavior or unintentional migration differences. If they were intentional then this PR can be closed, otherwise I think we should restore the status quo.

basil:
Subagent review:

> 1. auto: false on keyboard note/scale — confirmed against the original: updateMartyriaNote/updateMartyriaScale (old Editor.vue:6175-6190) passed auto: false, and f3e1521e's inlined keyboard path dropped it. The toolbar path was never affected because the note/scale selects only render when !element.auto. Your fix restores parity.
>  2. quantitativeNeume: null on alignRight change — confirmed: old updateMartyriaAlignRight (Editor.vue:6149) always passed it. Your centralized only-when-changed condition in updateMartyria is equivalent in practice (both the toolbar button and the keyboard toggle always flip the value) and covers both call paths.
>  3. ToolbarModeKey update:tempo — confirmed: the broken $emit('update', { tempo }) bypassed setModeKeyTempo, losing both the toggle-off-on-reselect behavior and the bpm recalculation from editorPreferences.getDefaultTempo(...). The telltale was that f3e1521e kept 'update:tempo' in emits and kept the parent listener but never emitted it. Your fix rewires it; the parent handler logic survived intact. No other emit in that file (or any other toolbar) has the same bug — the only other instance of this bug class was the ison button, fixed back in October by 1bf247e1 (#1759).

danielgarthur:
@basil I'm good with this PR going in.

PR #2256, opened 2026-06-10
fix: toaster causes blank page on print

### LINKED ISSUE #2252: still that extra blank page printing
Good morning, Daniel.  Sorry to say, I think there still is a bug with that printing an extra blank page, but only at times.  Last night, I exported two files, and they both had an extra blank page.  Then, this morning, the first one I did didn't have it.  But now the second one I'm exporting does have it.  I can assure you that that last martyria does not have a page break.  I'm attaching the PDF so you can see what's on the first page, and then the next blank page.  Let me know what you think...

Christopher

[Exapost-0807-10-After Transfig-BN.pdf](https://github.com/user-attachments/files/28800252/Exapost-0807-10-After.Transfig-BN.pdf)

danielgarthur:
Toast notifications created by Vue Sonner were appearing in printed output. Since the toaster renders its own notification region outside of the normal application content, it was not automatically excluded when printing.

This change hides the Sonner notification region when print media is active, preventing toast notifications from appearing in print previews and printed documents. The solution uses print-specific CSS rather than application state changes, which avoids timing issues with beforeprint events and Vue's asynchronous rendering cycle.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -2,7 +2,9 @@
   <TooltipProvider :delay-duration="500" :skip-delay-duration="0">
     <router-view />
   </TooltipProvider>
-  <Toaster />
+  <div class="toaster-wrapper contents">
+    <Toaster />
+  </div>
   <div v-if="updateExists" class="update-notification">
     An update is available.
     <button class="ok" @click="refreshApp">Update</button>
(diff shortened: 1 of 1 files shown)


basil (approved):
Sorry for introducing this and thanks for the fix.

PR #2279, opened 2026-06-15
feat: Custom toolbar for rich text editing

basil:
This PR replaces CKEditor 5's built-in balloon/UI toolbar with a custom Vue 3 + Reka-UI toolbar and properties panel, driving CKEditor's commands directly.

image

This provides a more unified and consistent UX while also giving us a more firm UI foundation on top of which to develop future editor features. New rich text editing features can now take advantage of either the toolbar or properties pane (with full shadcn-vue/Reka functionality in either case).

I ported over the custom neume insertion plugin. The buttons to insert custom neumes are now part of the rich text toolbar. Properties for these now live in the properties pane. I did not bother porting the CKEditor UI for link management, image upload, tables, or find/replace. These features are rarely used, and a non-native UI for them is not so much of a distraction.

The integration between CKEditor and Reka was far more difficult than anticipated. `notes/ckeditor_reka.md` describes the gory details around focus and selection management as well as flushing. While the complexity is high, I feel that the end result came out pretty well, with little to no dependency on CKEditor or Reka internals and no messy timing code. The result should be maintainable I think, as all the complexity follows a clear separation of concerns with structural enforcement of the layers as well as extensive documentation. I tested it interactively quite a bit to verify that there are no race conditions.

As a bonus, I enabled localization for CKEditor. This pulled in a few extra kilobytes of translation files, which increased the size of the bundle passed the threshold, so I had to bump the build to allow for a larger bundle size as well.


danielgarthur:
@basil I'm scared to look at the code, but the end result looks great.

basil:
Can you help with testing then? I've tested it quite a bit (and trust me, there were so many race conditions before I figured out the nuanced design that is documented carefully in the design document), and I can't seem to find anything broken, but I also don't use the feature all that much, so you might be able to exercise it more thoroughly.

danielgarthur:
@basil It seems actually less buggy than the original when it comes to the martyria editing, for example. I couldn't find any sure bugs. My only feedback is this:

- When you click on a martyria or neume, the `Properties` are not very discoverable. There used to be an in-your face popup, but now, if the properties menu is collapsed, you will never know that is there. I think maybe if you click the element, we should expand the `Properties`. I'm open to other ideas on how to make it obvious that there are options there. I would consider this a blocker on the PR.
- I somehow got the editor into a state where I could no longer insert bullets or numbered lists. This is not a blocker since I can't reproduce it and it may very well be a pre-existing issue. I only mention it in case you saw it too, then maybe it is worth some low-effort investigation. My guess is that it is a CKEditor issue and not some kind of break between the communication between the toolbar and CKEditor.

basil:
>  I think maybe if you click the element, we should expand the Properties.

Done in commit bb7f1d81ffc73baa9097769f8ce1942e1eae511b

I couldn't reproduce the issue with bullets or numbered lists.

basil:
> It seems actually less buggy than the original when it comes to the martyria editing

I noticed that too. For example, random issues I've had before, like saves not including data I just typed until after I leave the rich text box, seem fixed. I guess once I told Claude to systematically design a set of serialization invariants and then apply them to the whole codebase, it delivered.

danielgarthur (approved):

PR #2302, opened 2026-06-16
Add minimumSyllableToHyphenClearance

### LINKED ISSUE #2033: Need a setting for minimum gap between lyric and hyphen
This doesn't follow standard conventions:

Image

Note that there is an equal amount of space between the lyric and the first hyphen, the first hyphen and the second, and the last hyphen and the next word. Compare to Dorico with default settings:

Image

Note that between the lyric and the first hyphen (and the last hyphen and the next syllable) there is a smaller amount of space than the amount of space between two consecutive hyphens.

The relevant Dorico settings:

Image

I think the problem is that Neanes only has one of these two settings ("the minimum distance between adjacent hyphens", which corresponds to Dorico's "gap between hyphens" setting). Neanes should have a separate setting, the equivalent of Dorico's "minimum gap between lyric and hyphen" setting. And like Dorico, its default value should be 25% of the value of the other setting.

danielgarthur:
#2033


danielgarthur:
@basil Open question for you: do you think we should default PageSetup to something closer to Dorico's defaults? I think that is roughly:

hyphenSpacing = 0.525 in
syllable hyphen clearance = 0.13125 in

Those numbers are from AI, though and I don't have Dorico to confirm. I tried those values and personally found them more pleasing. What are you thoughts on best defaults?

basil:
Default space size is 1.75 mm:

image

So I believe that would lead to Dorico "gap between hyphens" of 13.125 mm and "Minimum gap between lyric and hyphen" of 3.5 mm.

I agree that we should use those defaults, Dorico is the gold standard here, and melismas in Byzantine music are, ahem, not uncommon.

basil (approved):
LGTM

PR #2303, opened 2026-06-17
feat: Clean up main toolbar

### LINKED ISSUE #2274: Clean up top/bottom permanent real estate
Perhaps the main toolbar should be more conventional, with buttons for New, Open, Save, Print, Cut, Copy, Paste, Undo, and Redo?

I personally feel that the "Insert/Remove Line Break/Page Break" buttons are too prominent. I almost never use these, and with the new Selection Pane in #2249 perhaps they don't need to be full-fledge main toolbar buttons.

Similarly, the "Page 1 of 1" seems awkwardly wedged at the top. I feel that it belongs at the bottom, in a fixed bottom toolbar, that can also display other "permanent" information, like the neume value (e.g. Ni - Di)? Something like this:

Image

(And yes, I realize I am dating myself with this screenshot.)

basil:
image

Fixes #2274


danielgarthur (dismissed):

danielgarthur:
@basil My only concern is that even with Page 1 of X removed, shrinking the window to the smallest size hides the playback section. A little space could be saved by moving zoom down to the status bar, too. See MuseScore, for example:

image

basil:
Good point. I changed the code so that it wraps to a third row on those small screens.

PR #2309, opened 2026-06-17
Show properties pane by default

danielgarthur:
@basil Any objections? Even on my tiny laptop screen (width is 11.7 in), the properties pane fits fine. My main concern is the discoverability of `Use Default Style` on text boxes, since that is a common thing to change.


basil (approved):
LGTM

PR #2314, opened 2026-06-18
Show lyric hyphen before first note of system

### LINKED ISSUE #2032: Show lyric hyphen before first note of system
By default, lyric hyphens should be shown before the first note of a system. From Dorico's default configuration:

Image

danielgarthur:
Closes #2032.

### DIFF
--- a/src/components/TheEditor.vue
+++ b/src/components/TheEditor.vue
@@ -17,6 +17,7 @@ import { useTranslation } from 'i18next-vue';
 import { debounce, throttle } from 'throttle-debounce';
 import {
   computed,
+  type CSSProperties,
   nextTick,
   onBeforeUnmount,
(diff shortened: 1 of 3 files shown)


basil:
This is causing some problems in my scores, not sure if it is an interaction with #2209 or an independent problem:

image

I'm going to revert for now so that I can cut a stable release and we can pick up where we left off later.

danielgarthur:
@basil I don't see that issue in that same score. However, I also don't have the font you are using. Is it still broken in latest `master`?

danielgarthur:
Okay, yeah, I was able to reproduce it before the revert and it is fixed now in `master`. 
image

basil:
All is good now. I will cut one more release and then rebuild my scores again.

PR #2319, opened 2026-06-18
feat: Agia attractions

### LINKED ISSUE #213: Diatonic Zo attraction algorithm is too simplistic

basil:
Redo of #2262. About 1/3 of the code, and far simpler, while still covering 95% of the cases of the original. @danielgarthur Are you OK with me renaming this feature from "Diatonic Zo Attraction" to "Agia Attraction", since it now covers both Zo and Ke? I did that in the code and docs in this PR, and also updated playback settings to include a box to customize the Ke attraction as well as the Zo one, so this is a first-class feature now.

Fixes https://github.com/neanes/neanes/issues/213


basil:
I think I simplified too much, now some Pl 1 scores are sharpening Ke too much. I'll take into draft and fix, then move back to non-draft status when I've tested more scores.

basil:
@danielgarthur Maybe you have some insight here, my algorithm is really close, and very simple, but it is slightly too simple for this case:

image

Any idea for how to fix that without drastically increasing complexity? I recall you were able to find some good tweaks to my ideas in the past, like for playback of scales.

basil:
OK, I fixed the biggest mistakes and checked all my scores, the result isn't 100% perfect, it's probably 85% to 90% perfect, but as far as I can tell, a strict improvement over the status quo without regression. More tweaking is certainly possible, if you'd like me to do that I can, it's effectively a judgment call as to when this is "done enough" for now (commit and iterate later) or continue the yak shave,,,

Issue #2281, opened 2026-06-15
Kerning


basil:
Collecting examples of kerning I've found in the wild.

basil:
In EZ Psaltica, the vareia is not as long as it is in classical books, perhaps due to lack of kerning?

<img width="1280" height="400" alt="Image" src="https://github.com/user-attachments/assets/7244725c-4c2e-414d-a624-8432b8e21f4b" />

In classical books, it is quite a bit longer, and almost always kerned.

basil:
The apostrophos is often kerned as well:

<img width="1036" height="394" alt="Image" src="https://github.com/user-attachments/assets/2abe3784-77ac-4e75-89ab-79ccd88c53f2" />

basil:
Apostrophos and kentimata can be kerned:

<img width="2218" height="686" alt="Image" src="https://github.com/user-attachments/assets/6f513e9f-ca87-4b41-8008-989f0816f3e9" />

basil:
Oligon and kentimata as well:

<img width="2174" height="752" alt="Image" src="https://github.com/user-attachments/assets/ff0b9e99-6c40-447b-8eca-5865f2ac0614" />

basil:
Some particularly tight pairs here:

<img width="3384" height="1402" alt="Image" src="https://github.com/user-attachments/assets/90a6a6b5-d6d6-4a89-a868-465eb2725a8c" />

basil:
<img width="3664" height="790" alt="Image" src="https://github.com/user-attachments/assets/5a966e10-d27b-4f7e-b395-e4e75007f33d" />

basil:
<img width="2086" height="1108" alt="Image" src="https://github.com/user-attachments/assets/c2929160-213a-4a42-95c5-b3160820365d" />

danielgarthur:
> Some particularly tight pairs here:
> 
[...]

PR #2323, opened 2026-06-19
Fix omalon spacing on oligonKentimaMiddle-ison boundary

### LINKED ISSUE #2079: Omalon too far to the left (oligon with kentima)
From the Kalophonic Heirmologion by Phokaeus:

Image

Neanes places this too far to the left:

Image

danielgarthur:
This PR adds font-driven note-pair kerning to layout so curated glyph pairs can tighten same-line note spacing without affecting saved score data or line breaks. Kerning is looked up by the notes’ primary quantitative glyph names and applied only in LayoutService.calculateInterNoteSpacing(...), where it still tightens the actual same-line base spacing. For note-to-note visual minimum spacing, primary base glyphs are no longer checked against each other, so curated base pairs can kern tighter without being pushed back apart by base-vs-base collision checks. Mark, vareia, inline-glyph, and measure-bar collision handling remain unchanged, and existing measure-bar spacing, martyria behavior, lyric collision handling, and non-kerned note pairs continue to behave as before.

image

Closes #2079


danielgarthur:
@basil Submitted for your consideration. This is the approach I am thinking about taking with kerning pairs in general.

```json
 "kerning": {
      "notePairs": [
        {
          "left": "oligonKentimaMiddle",
          "right": "ison",
          "adjustment": -0.13
        }
      ]
    }
```

basil (approved):
The visual result looks great. The only comment I have is that since this is generic infrastructure, it might be worth testing it with both tightening (the common case) and loosening (collision avoidance) kerning pairs to make sure both use cases are fully supported. The specific example tested in this PR seemed to be tightening only.

danielgarthur:
@basil Do you have an example of base neumes that need positive kerning?

Also, for many of the examples in #2281, they are not consistently applied, making me believe the general infrastructure needs to allow for increasing the allowed shrink, possibly in conjunction with some amount of negative kerning.

Compare

image

to

image

I'm not really sure what the default kerning should be here. How close are they by "default"?

I'm not sure how well this would work in practice, but perhaps we could use the OCR app on a PDF that has good spacing, then analyze the metrics. I believe the coordinates and widths are already available in the ocr file. If not they can be added.

basil:
I believe that kerning pairs in regular fonts are mostly tightening, but some have loosening for things like a lowercase f followed by a closing punctuation mark or quote, e.g. `f)`, `f"`, `f'`. I suppose for Byzantine music it could be used for collision avoidance in the base font (e.g. for ByzHtml, where the collision avoidance subsystem isn't in place) or to supplement the collision avoidance subsystem in special cases? There might not be a huge use case for this, but in any case it seemed like an angle to consider to make sure the general infrastructure is fully generalized.

Yeah, these old Greek books are full of inconsistencies, partly no doubt due to the fact that they were pioneering a brand new typographical form with limited budget. There is a lot that can be learned from them, but also we have to examine things critically under the light of the different economic and technological conditions we are in today. The interaction between bearings, kerning, and Knuth-Plass is indeed complicated, and I'm not sure what its final form should take exactly, but we're definitely moving in a good direction. The visual principles behind kerning are timeless. While I've never kerned a font myself, my understanding is that there are some [automated algorithms](https://patents.google.com/patent/US5803629) for doing it but they are widely hated by font designers, who prefer to define the metrics themselves, I guess because the algorithms often guess wrong. There's also this fun kerning game: https://type.method.ac/

basil:
A related tangent is the question of how many imperfections of the old books should be preserved vs flattened out. I actually have a wonderful book about this topic, _Revival Type_ by Paul Shaw, where he presents many examples of both philosophies. In short, some revivals are more of a "warts and all" approach, while others flatten and rationalize the font (sometimes too much, losing its character). I'm not a fan of the "warts and all" approach personally, and this is one of the reasons I love EZ Psaltica so much, because it does feel very rational and modern, while still having enough character to make it unique. The renowned font designer Robert Slimbach, who designed Adobe Garamond and Minion Pro, once wrote that (paraphrasing from memory) the digital medium, with its exactness and reproducibility, is more consistent and logical than the old print forms and therefore demands an increased level of logic and consistency from the font in modern times, which makes revival fonts particularly challenging. No doubt Slimbach has been one of the most successful modern type designers in achieving this balance.

danielgarthur:
I've decided to address this more narrowly for now with 8a79a9b. I have some ideas for kerning pairs, and kerning just that one thing all the time regardless of the omalon doesn't make sense, I don't think. The result is not terrible:

image

PR #2329, opened 2026-06-19
Add facing pages

### LINKED ISSUE #651: Support for facing pages
The [canons of page construction](https://en.wikipedia.org/wiki/Canons_of_page_construction) dictate that inside margins be smaller than outside margins. Suggest adding a **Facing Pages** checkbox to **Page Setup** which changes the **Left** and **Right** margin settings to **Inside** and **Outside** margin settings. In addition, the GUI should be changed to display two pages at a time when viewing a spread.

danielgarthur:
#651

image

Added facing-pages support without changing the saved horizontal margin fields. `PageSetup` now has a `facingPages` flag, a shared resolver maps `leftMargin`/`rightMargin` to physical page margins by page parity, and layout stamps each `Page` with a physical page number so editor rendering and export use the same source of truth.

The main behavior changes are:

- Layout now resolves per-page inside/outside margins for normal element placement and page-edge-sensitive placements.
- The editor flips guides, header/footer anchors, rules, and adjustment-ratio placement per physical page.
- LaTeX export now normalizes X positions against each page’s resolved physical left margin.
- HTML export emits `@page :left` / `@page :right` margins when facing pages are enabled.
- The Page Setup dialog adds a `Facing Pages` checkbox and relabels `Left/Right` as `Inside/Outside` when enabled.


basil (approved):
Manual testing worked fine for me, thank you very much!

The agents both independently flagged this issue:

---

| Case | LaTeX x produced | Correct (clean `xOffset`)? |
  |---|---|---|
  | LTR, non-facing | `xOffset` | ✅ |
  | LTR, facing (odd & even) | `xOffset` | ✅ (left resolves to inside/outside, and `element.x` tracks it) |
  | RTL, non-facing, symmetric margins | `xOffset` | ✅ (error term is zero) |
  | RTL, non-facing, **asymmetric** | `xOffset + (rightMargin − leftMargin)` | ❌ regressed vs master |
  | RTL, **facing** | odd: `xOffset + (outside − inside)`; even: `xOffset + (inside − outside)` | ❌ sign flips by page parity |

The RTL facing case is the worst: the error term flips sign between recto and verso, so it can't be fixed by any single global template offset. (Master happened to be clean for RTL because it anchored element.x at leftMargin for both directions and subtracted leftMargin — so master gave xOffset. This diff broke that coincidental correctness by re-anchoring RTL to the right edge in layout but leaving the exporter subtracting left.)

Fix is to anchor the exporter to the direction-appropriate edge, mirroring the editor:

```
  const contentAnchor = pageSetup.melkiteRtl
    ? resolvedMargins.right
    : resolvedMargins.left;
  // then each: x: toPt(element.x - contentAnchor)
```

That yields a clean xOffset for RTL in all four non-facing/facing × odd/even combinations and reproduces master exactly for the symmetric case.

danielgarthur:
My agents said the same thing, but I think they are wrong. RTL seems to work fine. It's probably because RTL is so janky that the agent just can't figure it out.

Issue #2276, opened 2026-06-13
Knuth-Plass line breaking algorithm regressions (continue from #2192)


ilizol:
Continue from issue https://github.com/neanes/neanes/issues/2192

"Knuth-Plass line breaking algorithm" PR: https://github.com/neanes/neanes/pull/2070

---

**Before "Knuth-Plass line breaking algorithm" (v0.5.8):**

[bug_line_break_v0.5.8.byzx] - [source](https://github.com/ilizol/neanes-examples/blob/5b3c281e26915c7219190d78366712bd14c156ea/bugs/bug_line_break_v0.5.8.byzx)
[bug_line_break_v0.5.8.png] - [source](https://github.com/ilizol/neanes-examples/blob/5b3c281e26915c7219190d78366712bd14c156ea/bugs/bug_line_break_v0.5.8.png)

<img width="1900" height="916" src="https://github.com/user-attachments/assets/fe9c5a09-0b4a-45b9-bda3-1cb5224153c5" />

---

**After "Knuth-Plass line breaking algorithm" (v0.5.25):**

- The 1st line has trailing space.

[bug_line_break_v0.5.25.byzx] - [source](https://github.com/ilizol/neanes-examples/blob/5b3c281e26915c7219190d78366712bd14c156ea/bugs/bug_line_break_v0.5.25.byzx)
[bug_line_break_v0.5.25.png] - [source](https://github.com/ilizol/neanes-examples/blob/5b3c281e26915c7219190d78366712bd14c156ea/bugs/bug_line_break_v0.5.25.png)

<img width="1900" height="919" src="https://github.com/user-attachments/assets/6984cd6a-1ce4-4363-bb3e-3888579d607d" />

danielgarthur:
What is the use case for this? I mostly consider `spaceAfter` dead, especially once #2209 goes in.

ilizol:
@danielgarthur 

This is a simple example that reproduces the line breaking algorithm issue, trailing space at the end of line.

https://github.com/neanes/neanes/issues/2192#issuecomment-4529514056

basil:
Fixed in #2333

PR #2333, opened 2026-06-21
Fix trailing space at line end

danielgarthur:
## Summary
- Move note-boundary justification elasticity from pre-break glue to post-break glue so positive adjustment ratio is not spent as invisible trailing line-end space.
- Keep martyria-owned note boundaries fixed so martyria spacing ownership remains unchanged.
- Update `knuth_plass.md` and inline comments to describe the new breakpoint encoding.

#2276


danielgarthur:
@basil @ilizol I asked GPT 5.5 to fix this and this is what it came up with. Seemed reasonable enough to me. I need to check more scores to make sure there is no collateral damage.

basil (dismissed):
LGTM, I didn't notice any problems in my scores when testing

basil (approved):

Issue #2251, opened 2026-06-10
Bookmarks


skatharis:
Hello everyone!

Lately I experimented with the hyperlink ability to interconnect different parts of a pdf.

Since this is already feasible, I thought it would be easy to add bookmarks using the same element tool in the program. This would help navigate long pdf files.

I am really happy to see the progress of Neanes and really gratefull, keep up the decent job!

basil:
Can likely insert these into a PDF with https://mupdf.readthedocs.io/en/latest/reference/javascript/types/OutlineIterator.html

PR #2350, opened 2026-06-24
Section breaks

danielgarthur:
## Summary

This PR adds Word-style document sections for headers and footers, using `pageBreak + sectionBreak` to start a new section on the following page. The feature now supports:

- Per-section header/footer content for sections after the first
- Per-section `Different First Page`
- Per-variant `Link to Previous` for `default`, `odd`, `even`, and `firstPage`
- Section-aware status bar numbering based on section breaks, not LaTeX `sectionName`
- Section-aware header/footer resolution in layout, editor selection, and save/load paths

## Behavior

- The existing Page Setup workflow remains the control surface for section 0 / document defaults.
- Later sections can be edited through the Selection pane while working in that section’s header/footer.
- `Link to Previous` is resolved per active header/footer variant, not at the whole-section level.
- Turning off `Link to Previous` materializes only the active variant into the current section.
- `Different First Page` is section-local for later sections and does not delete dormant first-page content when turned off.

## Backwards Compatibility

- Root `score.headers`, `score.footers`, and `score.pageSetup` remain in place for compatibility.
- `sectionName` is unchanged and still refers only to the LaTeX export section concept.
- Older scores without `sections` continue to behave as before.
- Section 0 still mirrors the existing root header/footer and Page Setup fields.

## Save / Undo Semantics

- Runtime section data is preserved while editing so undo/redo does not lose section-specific content or settings when section breaks are removed.
- JSON save now writes only sections that are currently active from the score’s section-break structure, rather than serializing dormant extra sections.

## Current Scope

This PR scopes section behavior to header/footer content and first-page/linking rules only. Most Page Setup options remain global, including things like odd/even mode, visibility, rules, and header/footer margins.

## Future Work

The resolver/model structure added here leaves room to move more page-setup features into sections later, including section-specific header/footer margins and other layout settings, without changing the basic section-break model or the existing section 0 compatibility story.


basil:
Very nice, will pair well with the new rich text features for small caps and old style numerals (often used in headers/footers). Some initial thoughts after playing with this:

- I see "Variant: Odd Page" after a section break, but not before a section break, do variants not apply there as well?
- The "Odd Page" part of the variant is inconsistently styled with the "Header" type right above it.
- The preposition "to" in "Link Header To Previous" is erroneously capitalized
- The "Line Break" (dropdown), Section Break (switch), and "Page Break" (button) all seem somewhat inconsistent, this was already a problem before this PR but this is definitely growing worse, might be time to think about how to refactor

As far as UX goes, I wouldn't consider Word's "Link to Previous" to be good UX, in fact quite the opposite, I think a lot of users are confused by it:

- The label is ambiguous, as “Link to Previous” sounds like navigation or hyperlinking, not “inherit this header/footer from the previous section”
- The state is easy to miss, it's hard to realize the current section is still linked until edits unexpectedly affect other pages
- The mental model is hidden, “previous” actually means previous section of the same header/footer type, which is not obvious
- The default is surprising, many users create section breaks specifically because they want something different, so defaulting to linked can feel hostile

Another downside to Word's UX is that while Word can auto-pull the current heading via a StyleRef field (`{ STYLEREF "Heading 1" }`), it's buried, so most people just type each section's header by hand, which is fragile.

I feel that a better UX design is one that focuses more on the semantics of the current section, a direction we are already going in via "Section Name (LaTeX)", the way that in TeX you define (with the `fancyhdr` package) `\fancyhead[L,C,R]{...} and \fancyfoot[...]{...}` (separate odd/even variants) using the running chapter/title/section once per document rather than hard-coding each new section.

I rather like the way InDesign does this with its Text Variables https://helpx.adobe.com/indesign/desktop/add-and-manage-text/conditional-and-variable-text/text-variables-overview.html

In fact just about any professional application (InDesign, FrameMaker, TeX) that is used for serious document processing models this semantically, unlike Word.

Ultimately I feel that our application should be moving toward more of a semantic model like TeX and InDesign rather than a model of hard-coding each individual section like Word. This would also allow us to generate PDF bookmarks more easily as requested in #2251, and also jives with our existing $p / $n variables and LaTeX section name functionality compared to the Word "Link to Previous" model.

danielgarthur:
@basil Thanks for the feedback. 

First of all, I will say that I am in favor of avoiding this large-scale section break PR if possible. But in order to do that, I want to make sure I understand what you are proposing.

Are you saying that we should simplify this by ditching the section break concept and instead adding, say, a `$s` context variable that would be replaced by the first `sectionName` that we find on the page (or the last `sectionName` in the document if the current page has no `sectionName`)?

I imagine that would address the main use case that users are probably try to accomplish. But it would not allow for completely different headers and footers for different sections. Although if users don't care about that right now and we can get away with the simpler option, I would prefer that.

basil:
> Are you saying that we should simplify this by ditching the section break concept and instead adding, say, a $s context variable

Yeah, the point is that rather than hard-coding unique data for each section, the header/footer would be a template filled in by the actual data, which is a less fragile and more maintainable document. For example, if you wanted to change the style of the headers, you would only have one copy to change rather than _n_ (classic Don't Repeat Yourself).

And one such data is encoded semantically, it can be used to generate a Table of Contents/PDF Bookmarks, another requested feature.

> that would be replaced by the first sectionName that we find on the page (or the last sectionName in the document if the current page has no sectionName)?

In terms of where the actual data is stored, the existing "Section Name" field is certainly good enough for a first version. This could possibly be refined further either in this PR or future ones, but it isn't essential. In InDesign you can define custom variables and map them to your document styles. That is probably overkill for this app, especially since we don't have support for paragraph/character styles, but the general idea that you might need to have more than one variable is sound (see my note below on use cases).

In terms of discoverability, in InDesign you can right-click and go to Insert Variable to see a list of them (or access this via a menu bar item). I think something similar would be useful here.

image

> I imagine that would address the main use case that users are probably try to accomplish. But it would not allow for completely different headers and footers for different sections. Although if users don't care about that right now and we can get away with the simpler option, I would prefer that.

As far as use cases go, I attach the relevant pages from Mitchell & Wightman's _Book Typography: A Designer's Manual._

IMG_2158
IMG_2160
IMG_2161


More than one variable is definitely needed for common use cases, typically (as the book describes) Author, Title, Chapter, and Sub-Title. In larger books, the main text may be divided into _parts_, each with a title page of its own, called a _part-title._ I'm not sure exactly how much you want to tackle in this first cut, but I think that given the software supports Facing Pages already, we should have at least 2 variables, one for the left side and one for the right side (see for example HTM's books). If a "Document Properties" metadata is in scope, Author and Title could be pulled from there.

One important use case (see the 3rd picture) is that folios and running heads or feet are not printed on

> Half-title, frontispiece, title page, imprint page, dedication/epigraph, contents pages, any pages before the contents page, part-titles, or any blank pages.

And

> And if folios and running heads are both placed at the top of the page, they should be removed above chapter openings.

The software's existing "different first page" is going in this direction, assuming the first page is some sort of chapter opening, but it is not flexible enough to meet true page layout needs.

danielgarthur:
@basil So before I spend the tokens, here's what I'm thinking. Your feedback is welcome. 

- Scrap the current PR.
- We add four running header roles:
  - Author
  - Title
  - Chapter
  - Subtitle
- Any element can be assigned one of these roles (not 100% sure about this one, maybe it should only be limited to text boxes? but I can think of edge cases where mode keys or images might be used)
- If an element is a text box element, the text in that box is used as the role value.
- There is an optional header role text field to override the text. For example, maybe you want an abbreviated version of the subtitle because the real title is really long.
- In header/footer you can use `$author`, `$title`, `$chapter`, or `$subtitle` to display the latest variable value. This is either the first value that appears on the page, or the running value prior. (not sure if it's worth it at this point to allow "last value on the page" as an optional switch).
- In addition, a document properties object in the UI should be added that allows you set the author and title. If no header role is present for author and title, the document properties are used. But header roles may be used to override. This could be useful for anthologies with different composers.

## Out of scope
These items are desired, but will not be part of this immediate scope of work.

- Document properties should be added as PDF metadata. 
- Header roles should be used to generate bookmarks.

basil:
Having Author and Title be overridable seems like unnecessary and avoidable complexity to me, the existing software I've seen makes these static fields. And your use case of anthologies with different composers could just be handled by using a different variable like Chapter or Subtitle.

---

While we're designing the object model, we should maybe think about some notion of level hierarchy and either (a) include it in the design, or (b) at least consider what it would look like to have it, and make sure our design doesn't preclude extension to levels later.

Here are the levels for TeX:

```
-1	\part{part}
0	\chapter{chapter}
1	\section{section}
2	\subsection{subsection}
3	\subsubsection{subsubsection}
4	\paragraph{paragraph}
5	\subparagraph{subparagraph}
```

When generating the bookmarks or Table of Contents these would be at different levels of the tree. Although Mitchell & Wightman use the term "Chapter" and "Subtitle", I find "Subtitle" a bit too generic and prefer the TeX terms "Chapter" and "Section" (which also jives well with our existing TeX "Section" UI). Probably people who are using this feature would have title on verso and chapter on recto, or chapter on verso and section on recto. But it should be obvious from the semantics and UI that these are not all the same level and have different effects on the ToC / bookmarks.

---

The only other thing that seems ambiguous to me is how (or if) we want to try to solve the problems mentioned about "When to delete folios and running heads". While the second example about running feet seems a bit obscure, the first one (removing the folio and running head from the first page of a chapter) is a common case. Here's an example of an otherwise well-typeset Byzantine music book I own that does this _incorrectly_ (the folio shouldn't be at the top right):

IMG_6116

I'm not sure if this book was typeset with Melodos or if Melodos supports this use case. But it should be possible in our software. Right now I don't see a way to do this easily with the current design. I don't think we need to make the design overcomplicated or tackle everything in one go, but we should at least think about what it might take to support this use case in the future and ensure we aren't precluding that in the current design.

danielgarthur:
@basil Revised plan below. Comments welcome as usual.

## Summary
Text boxes may optionally act as running markers: Chapter or Section. The displayed marker value comes from the text box contents unless an override is provided. Author and title live in Document Properties. Headers and footers can use `$:author`, `$:title`, `$:chapter`, and `$:section`. For each page, chapter/section resolve to the first marker on that page, or inherit from prior pages. A new chapter clears the inherited section until a section marker appears. Chapter-opening page options can hide page number, header, and/or footer.
## Details
- Add two running variable roles:
  - Chapter
  - Section
- Any text box element can be assigned one of these roles
	- The text in that box is used as the role value.
	- There is an optional running variable text field to override the text.
- Add a Document Properties dialog that lets you set author and title.
- In header/footer you can use `$:author`, `$:title`, `$:chapter`, or `$:section`. For each page, `$:chapter` and `$:section` resolve to the first marker on that page, if one exists; otherwise they inherit the most recent marker from prior pages.
- In Page Setup -> Headers and Footers, add options
	- Hide page number on chapter-opening pages
	- Hide header on chapter-opening pages
	- Hide footer on chapter-opening pages
- Chapter-opening page rules trigger only from chapter markers, not section markers. A chapter-opening page is any page where the resolved `$:chapter` value changes from the previous page’s resolved `$:chapter` value. A chapter marker resets `$:section` to empty until a section marker appears.
## Out of scope
These items are desired, but will not be part of this immediate scope of work.

- Document properties should be added as PDF metadata. 
- Header roles should be used to generate bookmarks.

basil:
That all sounds fine in general. This part still seems a bit ambiguous:

> In Page Setup -> Headers and Footers, add options
> Hide page number on chapter-opening pages
> Hide header on chapter-opening pages
> Hide footer on chapter-opening pages

The Mitchell & Wightman book says more precisely:

> If folios and running heads are both placed at the top of the page, they should be removed above chapter openings. If the folio is at the foot of the page, it should appear. Just the running head is removed. If you are using running feet containing chapter titles, you should remove these from the pages on which the chapter head appears [because the information does not need to be repeated]. The folio is left in its usual position.

So it's not necessarily the entire header and footer that has to be hidden on chapter-opening pages, but just the redundant running variable, which is hidden because it is redundant with the title. And the page number is only hidden on the top (but kept on the bottom) because otherwise it would be orphaned next to the white space above the title.

While some of this advice is subjective or a "best practice" rather than an absolute rule (and Mitchell & Wightman note that reference books and similar cases might require exceptions to the rule), I think the software should have good defaults, and should generally encourage good typography and book design through a simple UI experience, while making bad typography either a conscious user choice, or outright impossible.

In the end, I would just tell an AI the ideal rules for good design, described above, provide examples of common scenarios in my prompt and the desired outcomes (e.g. the examples from the Mitchell & Wightman book), tell it to keep things simple but make sure that all the common cases look good by default, tell it that exceptions should be possible as discussed above but rare enough that the user needs to opt into them rather than getting a bad default, and let it figure out the details...

PR #2356, opened 2026-06-25
fix: left-align leading bar lines

### LINKED ISSUE #2349: Should bar lines be aligned left
@basil When a bar line appears at the beginning of a line, it can be pushed right from the left margin when the lyrics are wider than the neume or when a hyphen is transferred. Is this the expected behavior or should the bar line be flush left against the margin, no matter how wide the lyrics are and how far the neume is from the margin?

Image

danielgarthur:
Closes #2349


basil (approved):
Thanks!

PR #2360, opened 2026-06-25
feat: add running variables

danielgarthur:
# Running Markers for Headers and Footers

## Summary
Add body-defined Chapter/Section running markers that can be consumed by any plain or rich text box via `$:chapter` and `$:section`, alongside document metadata tokens `$:author` and `$:title`. Keep the common book-typography behavior automatic: chapter-opening pages suppress top running matter by default, while bottom folios remain visible.

## Key Changes
- Add `Score.documentProperties` with `title` and `author`, saved as optional v1 score fields and edited through a new **File > Document Properties…** dialog.
- Add text/rich-text box marker fields:
  - `runningMarkerRole: null | 'chapter' | 'section'`
  - `runningMarkerText: string | null`
  - Only score-body text/rich-text boxes can define markers; headers/footers can consume tokens but cannot define marker state.
- Add a compact “Running Marker” section to `PropertiesTextBox.vue` and `PropertiesRichTextBox.vue` with role select and optional override field. If override is blank, derive the marker value from the box’s visible plain text.
- Extend `TokenMetadata` and `replaceTokens` for `$:author`, `$:title`, `$:chapter`, and `$:section`, preserving the existing `$p`, `$n`, `$f`, `$F` tokens.
- Add `PageSetup.useBookStyleChapterOpenings = true`, saved only when false, exposed in **Page Setup > Headers and Footers** as “Use book-style chapter opening pages”.

## Marker Resolution
- Build per-page running metadata from laid-out `filteredPages`, scanning pages in physical order.
- For each page, use the first Chapter marker and first Section marker on that page; otherwise inherit from prior pages.
- A Chapter marker updates chapter and clears the inherited section until a Section marker appears.
- A chapter-opening page is any page where resolved chapter changes from the previous page’s resolved chapter, including the first page with a chapter marker.
- Marker text extraction:
  - Plain text boxes: use the override when nonblank; otherwise read visible text fields in display order.
  - Rich text boxes: use the override when nonblank; otherwise strip HTML to plain text and normalize whitespace.
  - Empty marker values are ignored.

## Book-Style Opening Behavior
- When `useBookStyleChapterOpenings` is enabled and the page is a chapter-opening page:
  - In headers, `$:author`, `$:title`, `$:chapter`, `$:section`, and `$p` resolve to empty strings.
  - Header horizontal rules are suppressed on that page.
  - Footers keep `$p` visible, but `$:author`, `$:title`, `$:chapter`, and `$:section` resolve to empty strings.
- Existing first-page / odd-even header/footer selection remains unchanged.

## Test Plan
- Unit-test marker resolution across multiple pages: inherit values, first marker on page wins, chapter clears section, same chapter does not reopen.
- Unit-test token replacement for existing tokens and new `$:` tokens, including book-opening header vs footer behavior.
- Run `./node_modules/.bin/vue-tsc --noEmit` and `git diff --check`.

## Assumptions
- Document Properties lives under the File menu.
- Only body text/rich-text boxes define markers.
- The existing element-level `sectionName` remains separate and is not reused for running heads. It is not related to this task at all.


basil:
This looks good overall. It could use some documentation. I tried it out and it seems to work well.

---

When clearing the running foot from the bottom of a chapter page, I don't think this use case is currently handled correctly, although it might not be very common:

image

On the other hand, it's probably easy to fix with a regex capturing running foot and page number in the same box.

---

The LaTeX section name functionality feels orphaned both in the UI and feature-wise. I think it should be refactored to use section name from this new code rather than having its own (and could take advantage of chapter name as well).

---

The only use case gap I still see is blank pages. This probably isn't worth making the code very complicated to support, but I am mentioning it for completeness. Some house styles want new chapters to always start on a recto page and not a verso, but this is a subjective preference. If this house style is in place and the chapter would otherwise start on a verso page "naturally", then the verso page becomes a blank "spacer" page, with no header or footer at all and no content. This isn't needed if the chapter would "naturally" start on a recto anyway. (And, of course, all of this is inverted for Arabic books with a right binding.) And all of this only applies to Facing Pages, of course.

basil:
P.S. I still think a right-click menu or menu bar item for Insert Variable would help a lot in discoverability, and probably not too crazy to implement with shadcn-vue, but this is optional, not blocking:

image

danielgarthur:
@basil The case that isn't handled correctly: are you referring to the lonely dot if `$:section` is hidden but `$p` is not? And if so, the correct behavior is to show only the page number, right?

image

basil:
Yes, exactly.

basil:
Mark Argetsinger's _A Grammar of Typography_ has a different opinion from Mitchell and Wight, regarding folios on chapter pages:

> The page folio, if positioned in the running headline, should be placed at the foot of the page on chapter openings, rather than left out of account altogether.

And in his own book he moves the folio from the top left/right to the bottom (centered) on chapter pages. HTM does this as well. Since this is somewhat subjective, I wonder if the currently implemented `useBookStyleChapterOpenings` is too specific. It might be more flexible (and also more clear to the user) to implement this as "use different chapter opening" (like "use different odd/even") and have it be a completely different header/footer object with full customizability.

danielgarthur:
@basil I like that idea because then I don't have to do anything special regarding hiding stuff. Just let the user take care of it.

basil:
Agreed. I'd suggest the following defaults:

### Facing Pages (inverted for Arabic right binding)

#### Non-Chapter
Header:
Verso: Folio (L), Chapter (C), Blank (R)
Recto: Blank (L), Section (C), Folio (R)

Footer: Blank

#### Chapter (Argetsinger/HTM style)
Header: Blank
Footer: Blank (L), Folio (C), Blank (R)

---

### Non Facing Pages (TeX plain style)

Header: Blank
Footer: Blank (L), Folio (C), Blank (R)

danielgarthur:
@basil
- Added "use different chapter opening"
- Updated behavior for defaults

basil:
I tested this, the flow looks and feels sound. Of the potential remaining improvements, in the order of how strongly I feel about them:

- [Blocker] I think docs are essential for a new feature like this
- [High] An example score would probably help many users
- [Medium-High] I feel refactoring the LaTeX section name feature to use this metadata is somewhat important, as otherwise this is leaving behind object model technical debt in persistent on-disk state
- [Medium] Discoverability through menus would probably make the variables easier to find
- [Low/Out of scope] PDF integration of metadata

basil:
Also, since there's now a few different types of headers/footers (right, left, and chapter), it might be worth differentiating them in the UI somehow to help users not get lost.

danielgarthur:
By the way, there is an edge case with chapter-opening headers where the amount of reserved header/footer space can change after a chapter marker is discovered during layout. This is mitigated by caching the reservation on a per-page basis so layout remains consistent within that page, but it is not fully solved because doing so would require making multiple layout passes. In practice this only affects oversized headers or footers that extend beyond their allocated margins, which Neanes allows but are generally considered poor page setup. Really, I don't know that we should even support oversized headers/footers at all. But currently, we do and the layout processor will allot space for it.

danielgarthur:
I'm punting on this for now: Discoverability through menus would probably make the variables easier to find.

It's definitely a good idea, but text boxes don't currently have context menus at all, right now, and there's #2364. I think this should be incorporated into a larger context menu / usability push.

basil (approved):
LGTM

I thought it would be cool to do something with https://ckeditor.com/docs/ckeditor5/latest/features/merge-fields.html here, but alas, they are proprietary and not part of the GPL version.

basil:
> This is mitigated by caching the reservation on a per-page basis so layout remains consistent within that page, but it is not fully solved because doing so would require making multiple layout passes.

Yeah, pagination is a famously hard problem as well, and it's really tough to find a decent solution in one pass, let alone an optimal one. Michael Plass (the one from Knuth-Plass) wrote his [dissertation](https://tug.org/docs/plass/plass-thesis.pdf) on the topic and famously proved that the optimal solution is NP-complete, and TeX has used a one-pass greedy algorithm ever since. Although various people have come up with ideas for improvements over the years, Knuth famously refuses to break backwards compatibility in TeX, so none of them have ever been released.

danielgarthur:
One other thing: there is a breaking change. When loading old files with `sectionName`, the value is copied to the running variable only if the element was a text box or rich text box. Otherwise the data is lost. I consider this a scream test.

danielgarthur:
Example file will come separately of this PR.

PR #2370, opened 2026-06-26
feat: Text box over two pages

### LINKED ISSUE #1298: Text box over two pages
If a text box is at the bottom of a page and the text doesn't fit on the current page, it won't flow over to the next page. If I produce a pdf, there all the "extra" text gets truncated. The text is still visible in the editor, though it covers the bottom margin.

Would it be possible for a text box to automatically extend to the next page?

basil:
Fixes #1298

Body rich text boxes can now flow across page boundaries instead of overflowing a single page or forcing the user to split long passages by hand. A large rich text block — prose with headings, colored rubrics, justified paragraphs, and paragraph continuations — can begin partway down one page and continue onto automatically created continuation pages, while still behaving like one selectable, editable score element backed by a single HTML payload.

Previously, a non-scrollable box was a single block whose height was capped by the page. Long content had to be manually broken into multiple boxes, converted to a scrollable box, or accepted as an overflow problem in print and export. This PR adds a measured, line-boundary slicing pipeline so layout can keep the element as one box while rendering page-sized fragments in the editor, print, PDF, PNG, and LaTeX export paths.

Flow is limited to body rich text boxes whose line wrapping can be measured and replayed reliably. Inline lyric boxes, multipanel boxes, scrollable boxes, and center-on-page boxes keep their existing one-box behavior.

### Data model

The implementation separates persisted rich text data from transient layout geometry. `RichTextBoxElement` gains two unpersisted fields:

- `lineOffsets` — vertical cut candidates measured from the rendered content.
- `flowFragments` — the page-sized slices computed for the current layout pass.

A new `RichTextBoxFragment` records each slice's page position, content band, displayed height, layout height, index, and last-slice flag. It is a *view* into the source content, not a copied element. The durable fields (`content`, `languageCode`, direction, margins, width) remain the source of truth; the geometry fields are regenerated on each measurement/layout pass.

`Page.ts` gains a `richTextFragment` reference on `Line`. The origin slice keeps the element in the line's normal `elements` array (it's the real selectable element at its original location); continuation slices use `line.richTextFragment` instead of pushing another copy. This avoids duplicate elements, IDs, serialization state, and editing surfaces — there is still exactly one rich text box, with several layout fragments. The page-emptiness check was made tolerant of zero-line pages to fit the synthetic continuation-page flow.

### Layout and slicing

`LayoutService` clears every box's `flowFragments` at the start of `processPages` so stale fragments from a previous page/width/font/content state can't leak forward. After an element is assigned its page position and width, `createRichTextBoxFlowPlan` checks whether it's a flow-capable box, compares measured content height (from `lineOffsets`) to the remaining room on the origin page, and — when too tall — computes page-sized slices at measured line boundaries. The first fragment attaches to the current line; each continuation fragment gets its own synthetic `Page` and `Line` so it renders through the normal page template.

The slice math is split into two helpers:

- **`pickRichTextCut`** binary-searches the ascending offsets for the lowest legal cut: the largest offset that fits between the current content top and the available height. Returns `null` when even the first line below the top can't fit, giving the caller a clean "no cut here" signal.
- **`computeRichTextBoxSlices`** loops until the full content height is consumed, using the origin page's space for the first slice and a per-page callback for each continuation. It always appends the total content height as a final cut candidate, so trailing padding or a final block isn't silently dropped.

The per-page height callback matters because headers, footers, rules, first-page/odd/even variants, and running-marker state make available body height differ page to page. A shared "continuation height" is not precise enough when the first continuation page is shorter than later ones, so header/footer height is resolved per continuation page number.

Two edge cases are handled:

- **Fresh-page guard** — if a flow-capable box is first on a line, too tall for the remaining room, and can't fit even its first measured line, the service checks whether moving the line to a fresh page allows a legal cut, and moves it before slicing. This avoids a tiny/invalid origin slice in the last sliver of a page.
- **Forced progress** — if even a page-sized slice can't fit the next measured line (very tall rows, atomic/unbreakable content), the slicer takes at least one line so it can't loop forever.

Margins are accounted for in layout height, not the content band. The first fragment consumes its slice height plus the box's top margin; interior fragments consume only their band; the final fragment adds the bottom margin. This is why `RichTextBoxFragment` has both `height` (visible band) and `layoutHeight` (space the layout service reserves). Line-height calculation returns `layoutHeight` for flowed slices.

### DOM measurement

The browser is the only reliable source of rich text line breaks after CKEditor, font fallback, inline styles, neume/image widgets, tables, direction, and zoom. `TextBoxRich.vue` mounts a hidden, unzoomed, full-content data-view copy of the box whenever it's close enough to the page bottom to need offsets, or when an already-flowed box needs slice bands. The copy reuses the same `.ck-content` styling, width, font variables, direction, and child-margin normalization as the visible rendering, so offsets (for layout) and DOM ranges (for printed bands) come from the same document shape.

The hidden node is updated through `syncSliceMeasureContent` rather than `v-html`, because during editing the authoritative content is the live CKEditor data before `element.content` is flushed on blur; a template binding would compete and restore stale content. The component listens to CKEditor's `change:data` (via a new `data-change` event from `RichTextEditor.vue`), bumps a content version, and debounces remeasurement. On blur it refreshes offsets before emitting content updates, so the resulting layout pass slices the final breaks. Resize observers remeasure after fonts/layout settle.

Offset measurement walks the client rects of the hidden copy: it builds vertical bands from range rects, filters wrapper rects that merely contain inner line rects, clusters overlapping bands into rows, and records each row bottom as a legal cut (in unzoomed content coordinates). Atomic nodes — CKEditor widgets, inserted neume spans, image/table figures, images, tables, iframes, media, canvas, SVG, `<hr>` — are treated as indivisible: their bottom edge is a legal cut, and any candidate cut falling strictly inside one is rejected, so a boundary never splits a widget, image, table, or neume.

### Print-safe slice rendering

A tempting strategy — render the whole box in each fragment and clip to a page-sized viewport — is unreliable. Blink's print pagination can fragment an over-page block *before* ancestor overflow, clip-path, or paint containment clips it, leaking extra lines onto the next physical page or cutting glyph descenders, in PDF/print even when the clipped DOM looks right on screen. So print is never given a box taller than its page: each printed continuation slice contains only the HTML for its own line range.

Bands are produced by mapping measured fragment boundaries back to DOM ranges in the hidden copy. One mounted fragment owns the measurement copy and computes all bands for the shared element, cached in a `WeakMap` keyed by the element; other continuation fragments subscribe and render only their own band. The cache key covers everything that can change wrapping or visuals — measured HTML, width, direction, default font family, neume fallback font, default size, default color, and the fragment list — without making every fragment restringify on each render.

The boundary resolver handles text and non-text rows: for text it uses `Intl.Segmenter` (with a code-point fallback) to probe at grapheme boundaries and binary-searches for the first caret whose line top is at/after the requested boundary; for atomic rows it uses the node boundary before the element. It resolves many boundaries in one DOM-order pass and falls back to the root end. Bands are cloned with `Range.cloneContents`, preserving inline markup, styled spans, and widgets.

Cloned bands get two fixups when a cut falls inside a paragraph:

- A slice ending inside a justified block would make its last line lose justification (CSS "last line"), even though it's an interior line. The block is marked `rich-text-slice-continued-block` with `text-align-last: justify`.
- A slice beginning inside a block should not reapply the first-line indent. The first block is marked `rich-text-slice-continued-start-block` with `text-indent` reset.

These are scoped to split text boundaries and don't affect slices bounded by whole atomic blocks.

### Visible rendering modes

A supported body box renders in one of three ways:

- **No fragment** — the existing single editor, with max height adjusted for page-bottom space and footer-overflow reservation.
- **Origin fragment, editing** — the full-height editor stays mounted inside a slice-height container that scrolls internally while focused, so all source content stays editable even though only the origin-page window shows.
- **Continuation fragment (or origin in print)** — the precomputed read-only band HTML.

Focus handling combines raw DOM focus with the active-editor registry so focus moving into toolbar/portal chrome still counts as editing. When focus truly leaves, the container resets `scrollTop` to 0 before returning to the clipped origin view, preventing the first page from showing a scrolled band after overflow flips from `auto` back to `hidden`. The focused origin slice is widened by the platform scrollbar width when classic scrollbars consume layout space, keeping the scrollbar in a gutter.

### Editor orchestration

- **`RichTextEditor.vue`** subscribes to `editor.model.document` `change:data` and emits `data-change`, unregistering on destroy alongside the existing listeners. Only the single flow-capable body editor uses it; the existing model-value echo guard still prevents redundant `editor.data.set()` selection resets. This keeps offsets current during editing instead of waiting for blur.
- **`TheEditor.vue`** makes rich text recalculation awaitable. The old counter (how many boxes changed height) was insufficient, since a box may finish with unchanged height, changed/unchanged offsets, or multiple resize reports. A new `richTextBoxCalculationDone` set records element IDs that emitted a `measured` event during the current pass; `recalculateRichTextBoxHeights` resolves after every expected box reports completion (or after the existing timeout).
- The active-workspace watcher triggers measurement when a workspace becomes active. Because `lineOffsets` are transient, a score restored from local storage, opened from disk, or switched to may have content/heights but no offsets; without eager measurement a tall off-screen box (under virtualization) could go unmeasured, making page counts wrong. The watcher is the common point for open/restore/tab-switch.
- A `suppressActivationMeasure` flag avoids races during silent export: silent PDF/HTML/LaTeX loops switch the active workspace but immediately run their own awaited measurement, so the activation watcher's fire-and-forget pass is suppressed across those loops and lowered before interactive multi-file opening. (Interactive multi-open coalesces into one activation callback for the final workspace; others measure on later switch.)

### Export paths

Print, PDF, PNG, and LaTeX now explicitly await measurement before consuming pages: `onFileMenuPrint` and `onFileMenuExportAsPdf` measure and save before entering print mode; PNG measures before using the page DOM; LaTeX measures before passing `pages.value` to the exporter; the silent LaTeX path switched from `nextTick` to the awaited recalculation. This ensures the transient geometry exists and is applied before any consumer reads the page list — it does not persist offsets.

Page-bottom space for measurement now accounts for footer-overflow reservation, since footers and footer rules can extend past the configured bottom margin. `TheEditor.vue` computes the footer-rule reserved height, per-page footer reservation, and the max reservation across variants; live rendering passes the current page's reservation to each `TextBoxRich`, and the hidden recalc template uses the max (conservative before a specific page is known). `TextBoxRich` subtracts this when computing available height.

Page-setup change detection was widened from font defaults to include `innerPageWidth`, because changing page width or horizontal margins changes wrapping even when fonts are unchanged. Undo/redo/page-setup updates compare the richer snapshot and recalculate when wrapping width changes, keeping `lineOffsets` (a function of width, font metrics, and direction) from going stale.

### RTL and zoom

Continuation positioning uses each physical page's resolved margins and chooses the right margin or content-left start by `melkiteRtl`, matching existing element positioning; `TheEditor.vue` positions fragments with `left`/`right` via the same `rtl` flag. `TextBoxRich.vue` applies measured content direction to both the hidden copy and visible bands. The empty-content default still chooses Arabic RTL under Melkite RTL / eastern Arabic numeral settings.

The hidden copy is unzoomed and off screen; measured rects are normalized to its own top and returned as plain pixel offsets, while visible positions are scaled with `withZoom` at render time. Editable height measurement now targets `.ck-editor__editable` specifically — a broad `.ck-content` query would read the hidden unzoomed copy and report a scaled height.

## Limitations

Flow is intentionally disabled for:

- **Inline boxes** — tied to lyric layout with separate top/bottom editors; different measurement/positioning rules.
- **Multipanel boxes** — independent left/center/right regions; a single linear offset sequence wouldn't describe them.
- **Scrollable boxes** — meant to remain one fixed viewport with internal scroll.
- **Center-on-page boxes** — use an editor right-padding trick, so editable wrapping width differs from the read-only continuation width; cut points could disagree with visible breaks.

These keep the old one-box behavior, and line offsets are cleared when a box no longer qualifies.

## Tradeoffs

- Offsets are measured/cached transiently rather than serialized — saved files stay stable, but opened files need an eager measurement pass before layout is final.
- Correctness over coverage: center-on-page and multipanel flow can be revisited once the renderer guarantees identical wrapping.
- The band renderer uses DOM range probing + cache ownership rather than CSS clip, because clip was specifically invalidated by print behavior.
- The origin slice scrolls internally while focused — a pragmatic editing compromise; printed/unfocused display stays true to page slices.

## Testing

**Automated** — `LayoutService.test.ts` covers the deterministic slicing helpers (no DOM): largest legal cut within space, `null` when the first line can't fit, ignoring offsets at/below the current top, single slice when everything fits, two-page and three-page cuts, per-page available space, forced progress for an over-tall first line, a single unbreakable line as one slice, and appending total height when offsets stop short.

**Manual** — opened the updated `examples/Rich Text.byzx` (a long Greek passage previously stored as two boxes, now one) and confirmed it flows as one selected box; edited near the start and middle and verify page count and continuation content update after blur and during remeasure; exported to PDF/PNG and confirm no lines leak across boundaries.

## Compatibility

The byzx schema is unchanged: a saved file still contains one rich text element with normal formatting properties. On reopen, the editor remeasures and reconstructs fragments from current page setup, fonts, and DOM layout. Old files stay compatible, no browser-specific measurements are persisted, and exporters stay insulated — they consume pages after layout produces fragments, not a new durable fragment model.

## Contract

The key contract is between the measurement component and the layout service:

- **`TextBoxRich`** guarantees `lineOffsets` are ascending, unzoomed content-bottom positions where a cut won't split a row or atomic object.
- **`LayoutService`** cuts only at those offsets, includes the total height, consumes at least one row when no legal fit exists, accounts for per-page header/footer space, and attaches fragments without duplicating persistent elements.
- **`TheEditor`** runs the measurement pass before any page-consuming export and passes the right fragment and footer-reservation props to renderers.


danielgarthur:
I've seen some odd behavior I think relating to one fundamental bug. But I cannot reliably produce it. I think the issue is that the content split across pages does not always refresh when I expect it to. I will keep playing with it to give more concrete details.

basil:
My gut feeling is that it "should" be cleaned up more, but I'm not sure how...

basil:
> I've seen some odd behavior I think relating to one fundamental bug. But I cannot reliably produce it. I think the issue is that the content split across pages does not always refresh when I expect it to. I will keep playing with it to give more concrete details.

@danielgarthur I did a Fable 5 cleanup and bug-hunting pass and it claims to have fixed this and a few other Vue reactivity issues, please check again.

basil:
Performance/efficiency notes below. I did several optimization passes over the code looking for quadratic algorithms. GPT 5.5 was able to get everything down to linearithmic O(n log n) time with a few sort operations, a suffix minimum array, and use of the [sweep line algorithm](https://beyondctci.dev/editor?problem=company-launches).

---

The implementation improves efficiency by turning rich text pagination into a small set of reusable measurements. It measures the rendered content once, records each valid line-bottom cut point in a sorted `lineOffsets` array, and feeds those offsets back into layout. The layout service then chooses the largest cut that fits a page with binary search, rather than walking every line for every slice. Because the offsets are absolute content coordinates, each slice can be represented by only `offsetTop` and `height`. That keeps continuation planning compact even when a text box spans multiple pages. The result is page splitting work that scales with the number of measured line boundaries and slices, not with repeated full DOM layout passes.

The DOM band processing uses similar one-pass structures to avoid quadratic checks. `removeContainerFlowBands()` builds a suffix-minimum array of later band bottoms, so it can tell whether a rectangle contains a later inner rectangle without comparing every pair. Atomic content such as images, tables, and CKEditor widgets is handled by merging sorted cut candidates with sorted atomic bands using one moving pointer. Boundary requests for rendered slices are sorted and resolved during a single DOM-order traversal, with binary search used only inside text nodes that actually contain a requested line. The rendered band HTML is also cached in a shared `WeakMap`, so one mounted fragment computes the split content and the other fragments reuse it. These choices reduce repeated measuring and cloning while preserving precise line-safe cuts for screen and PDF output.

basil:
@danielgarthur Gentle ping, I think I'm done working on this code and testing it, so if you want to do more testing now would be a good time. If you have any concerns about the complexity of the code or the design, I'm happy to rework anything, or if you don't want to go in this direction I'm happy to close this PR.

danielgarthur:
I have begun looking at this.

PR #2371, opened 2026-06-26
fix: Line spacing for small font sizes in rich text boxes

### LINKED ISSUE #2365: Line spacing for small font sizes in rich text boxes
Discovered while trying to convert my footers to rich text boxes: CKEditor sizes inline spans but leaves the parent block at the editor's default line-height, so small inline text still reserves a default size line box. As such, my two lines of 8pt text in the footer were each getting a 12pt line height.

danielgarthur:
## Summary

- Improve rich text box font-size floor detection so editors can shrink below the Page Setup default when saved content uses smaller font sizes.
- Apply that derived minimum before hidden recalc measurements, including multipanel text boxes, so height recalculation uses the real rendered size.
- Keep mounted editors in sync when Page Setup defaults change, and correctly normalize both `pt` and `px` rich-text font-size attributes when computing the minimum.

Closes #2365.

### DIFF
--- a/src/components/TextBoxRich.vue
+++ b/src/components/TextBoxRich.vue
@@ -114,7 +114,14 @@
 import type { Editor, EditorConfig, FontSizeOption } from 'ckeditor5';
 import { debounce, throttle } from 'throttle-debounce';
 import type { PropType, StyleValue } from 'vue';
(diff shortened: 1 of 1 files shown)


basil:
Looks great, thank you! Now I can finally use old-style numerals and true-drawn small capitals in my footers:

image

basil (approved):

Issue #2382, opened 2026-06-27
Playwright tests


basil:
I believe https://playwright.dev is the modern alternative to Selenium for frontend testing. At least that is what my agents reach to when they need to drive a GUI to manually test something. The agents seem to do OK fixing real bugs by firing up the app in Chrome with Playwright and clicking on random buttons to do stuff. However, because there is no existing end-to-end frontend testing in the repo, the agents waste a lot of time writing throwaway Playwright code while doing their manual testing. I've watched them consume dozens of turns figuring out basic stuff like how to insert neumes, edit text boxes, and the like. They eventually figure it out, but they don't remember it (unless I use agent memory systems and the like). I think a concerted effort to build out end-to-end tests in Playwright would not just help in ensuring that we ship releases without regressions, but having even just a few of these tests written and checked in to the repo would be enormously helpful to agents who are trying to figure out how to drive the app when manually testing some new feature.

PR #2400, opened 2026-06-30
feat: Add paragraph styles

### LINKED ISSUE #1283: User-defined styles
Users should be able to define text box styles and given them names, then apply the style to a text box by choosing the style from a dropdown. Styles would be defined by configuring an existing text box and clicking a "save as style" button or in some similar fashion.

Branched from #1259.

danielgarthur:
Closes #1283.

## Summary

This PR replaces the old page-setup-driven text defaults with document-level Text Styles. Text boxes, rich text paragraphs, annotations, lyrics, drop caps, headers, footers, layout, and exporters now resolve typography through `Score.textStyles` and per-element overrides.

## Motivation

Text styling was split across Page Setup defaults, object-level override fields, and CKEditor paragraph state. That made defaults hard to reason about, made rich text behave differently from regular text, and complicated export/load compatibility. This change centralizes reusable text defaults into named Text Styles while preserving legacy files through load-time migration.

## Major Changes

- Adds document-level Text Styles with built-ins such as `Default Text`, `Annotation`, `Lyrics`, and `Drop Cap`.
- Adds Text Style management UI and wires existing text controls to resolved style inheritance plus per-object overrides.
- Integrates CKEditor paragraph styles so rich text boxes and annotations can apply Text Styles to selected paragraphs.
- Moves lyric, drop cap, text box, header/footer, annotation, layout, LaTeX, and ByzHTML defaults to resolved Text Styles instead of live Page Setup fields.
- Keeps regular and rich text box spacing in code as `marginTop` / `marginBottom`, while the properties UI labels those controls as “Gap Above” / “Gap Below”.
- Adds File menu access for Text Styles and bridges Page Setup sections to the Text Styles dialog.

## Breaking Changes

- New saves write `textStyles` and `textStyleId` fields. Older app versions may not fully understand the new style model.
- Legacy Page Setup text default fields are now migration inputs rather than the live source of truth.

## Implementation Notes

- Built-in styles are seeded from legacy Page Setup values on load, then `Score.textStyles` becomes authoritative.
- Element-level fields now represent explicit overrides; `null` means inherit from the selected Text Style.
- Rich text does not persist a single object-level Text Style ID; it uses CKEditor paragraph classes plus fallback style resolution.
- Exporters compare resolved styles against built-in defaults so inherited content stays compact while explicit/non-default styling is preserved.
- Margin storage remains `marginTop` / `marginBottom` for text boxes and rich text boxes, despite the gap-oriented UI labels.

## Reviewer Notes

- Save/load compatibility for old files with legacy Page Setup text defaults, sparse style arrays, and old `useDefaultStyle` behavior.
- Rich text style selection, especially focus handoff between editor, toolbar, and properties pane.
- Text Style deletion/fallback behavior for text boxes, rich text paragraphs, annotations, lyrics, and drop caps.
- Export parity in LaTeX and ByzHTML for inherited versus explicitly overridden typography.
- Layout changes around lyric measurement, drop caps, inline text boxes, headers, and footers.
- The properties panes’ handling of inherited values versus explicit overrides.

## Testing

- This PR adds/updates tests for Text Style resolution, SaveService migration/round-tripping, rich text style command state, ByzHTML export behavior, and rich text language handling.

## Known Limitations

- None known.


basil (commented):
Very nice, this is a great step toward stabilizing the on-disk format and making it possible to author more maintainable documents. I will definitely use this, since I have had to refactor my styles multiple times and it has been a pain to use AI to do global search and replace of styles. Here are some of my thoughts after trying this out briefly this morning:

- When creating a new custom style, I can't clear the name to type my own name. Every time I clear it, it gets repopulated with "Text Style"
- shadcn-vue vertical orientation seems inconsistent with the existing page setup dialog and properties pane
- The override switches are a great UI for expressing inheritance but they look a bit busy on the right next to the value. In Dorico they're on the left next to the key. I'm not sure this would necessarily look better but it might be worth playing with.
- In Dorico, styles without a parent lose their override sliders since there is nothing to override, this might make sense here as well.
- The create/clone/delete buttons look a bit plain, this could probably be touched up / facelifted somehow, I'm not sure how (icons? Reka UI button group? something else?). In general the visual style looks a bit unpolished compared to the other dialogs. I usually find that "touch up / facelift" style prompts work well, followed by (human) visual inspection and then complaining until it looks right. Another prompt I use often is "which shadcn-vue overrides are pointless (and can therefore be deleted to pick up shadcn-vue defaults) vs which ones are truly necessary in this dialog?"
- The dialog uses rounded corners which is incompatible with the general design language of the Lyra theme used in the rest of the app which uses square corners
- The text "Create reusable text-box text styles and control inheritance between them" sounds like AI slop and not a real user-friendly description. Ordinary users don't know what inheritance is in the context of computer science
- Dorico displays a * next to dirty unsaved styles and offers a style-wide Reset button, InDesign similar has a "Reset to Base" feature, this seems useful
- "Text Styles" seems a bit too basic, there is usually a concept of "Paragraph Styles" vs "Character Styles" in most such UIs, including InDesign, Dorico, even CKEditor itself has two separate selects ("Block Styles" vs "Text Styles"). This feature seems more like a "Paragraph Styles" feature. Even if a "Character Styles" feature is out of scope for this PR, we should probably design for extension in the nomenclature
- This type of feature usually is never in the File menu. Examples I've seen are Format (Word), Type (InDesign), Library (Dorico), but never File
- Design of the left-hand side is inconsistent with other dialogs, not sure a two-row "Title / Built-in" structure is necessary, can probably simplify/cleanup to the same design language used in other dialog boxes
- "Tombstone" dead UIs in page setup forwarding to the new UI generally seems like poor UX and an anti-pattern in product design. While it ensures users know the old UI is gone, it creates friction by forcing an extra click, infantilizes users, confuses new users who haven't seen the old UI, introduces nested modal dialogs (undesirable), and creates technical debt in the form of an unfinished migration

danielgarthur:
@basil I believe I have addressed all the feedback (which was very much appreciated) except this

> Dorico displays a * next to dirty unsaved styles and offers a style-wide Reset button, InDesign similar has a "Reset to Base" feature, this seems useful.

I wasn't sure exactly what this meant. I did however include a "Reset to Factory" that resets built-in styles back to their factory defaults.

basil:
By that I meant that "dirty" styles were marked, which helps when you're editing a lot of them and can't remember which ones you've modified (optional improvement, not essential):

image

Regarding the reset thing, I think it's essential that some sort of Reset button exist, because toggling many individual switches is too annoying. In Dorico it's a Reset button that just resets the current style (as opposed to resetting all built-in styles to factory):

image

In InDesign it also just resets the current style and is called "Reset to Base". (I haven't looked at Word.)

image

danielgarthur:
@basil I may have already done the reset thing. When you click it resets the selected style to the built-in default for that style. I assume that is what you mean as opposed to literally toggling all override switches to off.

basil:
I took another look at this, it looks much better.

- The behavior in plain text boxes of bold/italic somehow being different from underline now seems like a wart that can be cleaned up, I think the original intention of this was that underline was not possible to set in the page setup options, but was possible to use from plain text boxes. Now that this subsystem exists the three should all be first class citizens that can all be set or overridden at will. (Anyone who is using underline in professional music typesetting is doing it wrong.)

The current branch is a bit of a mess with regard to "Remove Format" or "Reset to Default" especially after merging in some recent changes of mine, where I tried to purge all Reka `<Button>` elements from properties panes. Here is my current thinking:

- Full-width Reka buttons (like the "Remove Format" to remove all formatting) don't belong in panes, and should be toolbar buttons. I already moved most of these, like neume positioning. Any remaining ones should be moved over.
- The "reset to default" buttons aren't full width and are an exception to the above, since they apply to individual styles and not all styles, these seem fine to keep in the pane, but I have struggled to find a good visual design for them. Maybe with our recent changes to do a proper design for the Paragraph Styles dialog, we want to adopt the same design language here, with them being on the left, and a circular arrow icon instead of an eraser.
- Apart from the button to clear all formatting, I don't think any individual "reset to default' buttons should be placed on the bottom toolbar, for either plain text, rich text, or any other bottom toolbars, as this is more of a fine-grained option and would otherwise clutter up the bottom toolbar (and is a more niche / advanced use case anyway).

Another lingering question is how to integrate this with the running marker subsystem. Currently you have to set the Chapter/Section paragraph style, and then also set the running marker role to Chapter/Section, which is not ideal. In InDesign the way this works is that the running variable is actually picked up from the items that have a given paragraph (or character style), rather than the whole text box. So you don't need to apply the role separately at all, but just apply paragraph styles as you normally do, and then these become variables that you can use in running headers/footers. To me this is a better UX than the current system where the entire box has to be used to define the variable.

I still haven't reviewed the code or persistence model, but I will take a look at that tomorrow.

basil:
One other refactoring to consider, currently not all surfaces can apply all styles, for example only rich text boxes can apply true-drawn small caps and old style numerals. I think plain text boxes had some limitations around underline (mentioned in my previous comment), etc. Ideally all these surfaces would support the same set of options, although I stopped short of this in my previous work because it was a huge amount of effort, and some of the combinations didn't make sense (e.g. you'd never need to right-align or underline a drop cap). But I think there is still some low-hanging fruit left on the table (e.g. small caps in plain text boxes, or ligatures in lyrics) -- if the effort isn't too high and it doesn't cause a huge amount of unrelated code churn, it might be worth expanding some of these surfaces opportunistically to support the same set of options.

danielgarthur:
@basil I've made "Text Decorations" separate and removed the "Remove Format" button. Beyond that, I think the other feedback should be part of separate PRs since this is getting larger than I'm comfortable with.

danielgarthur:
I think we just cleaned up the same thing...

danielgarthur:
Seems like your changes did the same thing as mine, so I guess no issue.

danielgarthur:
@basil FYI, CKEditor supports character styles. Currently paragraph styles are defined as follows.

```typescript
const paragraphStyleDefinitions = computed(() =>
  props.paragraphStyles.map((style) => ({
    name: style.id,
    element: 'p',
    classes: [`neanes-style-${style.id}`],
  })),
);
```

Character Styles could be defined using `span` as the element.

```typescript
const characterStyleDefinitions = computed(() =>
  props.characterStyles.map((style) => ({
    name: style.id,
    element: 'span',
    classes: [`neanes-style-${style.id}`],
  })),
);
```

I don't plan on making that part of this PR, but I wanted to note it for future work.

basil:
Sorry for the conflicts. I worried I had exhausted you with my review feedback, so I decided to help out and write some code. I just pushed a change that attempts to define consistent taxonomy/terminology:

- "Clear formatting" means "Reset override" and go back to base style (PhTextTSlash icon)
- "Reset Style to Default" means reset the definition of the built-in style to its factory default

As part of this I made the Format Paragraph Styles dialog support both use cases, showing "Clear formatting" for styles that have a parent, and "Reset Style to Default" for built-in styles.

I also moved all migration logic to the boundary so that the common code path for new documents and migrated documents doesn't execute legacy migration code. That seemed like a general code quality improvement.

I also made it so that migrated styles and new default documents didn't override anything with the same value. That makes sense to me as default behavior. If users want to override a value with the same value after the fact, we preserve that.

I'm passing the ball back to you, as I won't have any more time to work on this today. Things I didn't have time to investigate:

- I believe there is a dead Vue handler/target of some sort in Page Setup from the tombstone removal
- Searching for fonts doesn't seem to work for me in the Format Paragraph Styles dialog. Some sort of focus issue when I try to search
- AI-assisted audit of every setting exposed in paragraph styles to make sure it is fully plumbed through and made to take effect in all surfaces. I believe there is at least one bug here (line height not taking effect for lyrics)

danielgarthur:
Is it intentional that Rich Text Box no longer has a button that resets a selection back to its current paragraph style? E.g. if I highlight everything in the box and set the style to "Title", but then change the font size, I can revert the font size change by pressing the "Clear Formatting" button next to font size. But if I make multiple changes, I have to revert each individually. Plain Text Boxes, on the other hand, still have the "Clear Formatting" button next to "Edit Styles", which reverts everything back to "Title".

There is a "Clear Formatting" button in the toolbar, but this removes the Paragraph Style, too.

danielgarthur:
> Searching for fonts doesn't seem to work for me in the Format Paragraph Styles dialog. Some sort of focus issue when I try to search

I was not able to reproduce this on macOS, but I was able to reproduce it on Ubuntu. Odd.

basil:
> Is it intentional that Rich Text Box no longer has a button that resets a selection back to its current paragraph style? E.g. if I highlight everything in the box and set the style to "Title", but then change the font size, I can revert the font size change by pressing the "Clear Formatting" button next to font size. But if I make multiple changes, I have to revert each individually. Plain Text Boxes, on the other hand, still have the "Clear Formatting" button next to "Edit Styles", which reverts everything back to "Title".

Ah good point, this wasn't intentional and I was just overzealous when trying to unify the taxonomy/terminology. I think you're right that this should match plain text boxes.

basil:
image

This is not ideal. I understand if we want to keep the scope of this PR small, but we should ideally fix this pretty soon after this PR if so

basil:
As far as UIs for Bold/Italic/Underline go, there are basically two paradigms: the "consumer app" paradigm, where B/I/U are always next to each other, and the "professional app" paradigm (where underline is rarely used). I like how InDesign manages to put 6 buttons in a tight space, and aligns All Caps/Small Caps, Superscript/Subscript, and Underline/Strikethrough in this default toolbar:

image

basil:
It occurs to me that the simplest solution for our UI might be to remove the B/I buttons from the properties pane but keep them in the bottom toolbar. So the bottom toolbar can have the abbreviated consumer B/I/U pattern as 3 buttons, while the properties pane (for both rich text and plain text) can be a more pro-oriented view, with only a style dropdown for Bold/Italic (like the InDesign screenshot above) and buttons for underline, small caps, super/subscript, etc.

danielgarthur:
That could work. 

The problem I was trying to solve there was "how do I have a `Clear Formatting button` that makes sense?" That was just a quick way to make it work. But I also had the thought that the style buttons should be removed in the property pane.

basil:
I don't have a strong preference, but I think rich and text boxes should ideally be consistent in the UI. Right now rich text boxes show this, while plain text boxes have B/I separate from U, which is strange:

image

danielgarthur:
> I believe there is a dead Vue handler/target of some sort in Page Setup from the tombstone removal
Searching for fonts doesn't seem to work for me in the Format Paragraph Styles dialog. Some sort of focus issue when I try to search

Cleaned up.

> AI-assisted audit of every setting exposed in paragraph styles to make sure it is fully plumbed through and made to take effect in all surfaces. I believe there is at least one bug here (line height not taking effect for lyrics)

Line height intentionally is not set from the paragraph style. It is instead calculated via `TextMetrics` so that everything lines up correctly.

basil:
commit 5445421f708809e377d5e8de189106ac2c4843f9 doesn't seem to work for me, as I still can't search for fonts. If I have time tomorrow I will try to debug.

I tried the audit I mentioned above and got these results:

1. Some text formatting values can be changed on an element, but do not survive every element data path. Part of this is pre-existing: `origin/master` already has `lyricsTextDecoration` and text-box `lineHeight`, but lyric clipboard data does not include `lyricsTextDecoration`, and text-box clipboard data does not include `lineHeight`. This branch makes the problem worse: paragraph styles add `ParagraphStyleOverrides.textDecoration`, but element-level underline fields are not fully translated into that resolver shape. As a result, copied lyrics can lose underline state, including explicit underline clears, and copied text boxes can lose line-height overrides. Paragraph-style resolution can also miss explicit underline clears on lyrics or text boxes. The fix is to add `lyricsTextDecoration` to `NoteElement.getClipboardProperties()` and `lineHeight` to `TextBoxElement.getClipboardProperties()`, and to map lyric `lyricsTextDecoration` and text-box underline values into paragraph-style text-decoration overrides, using `null` to preserve explicit clears.
2. Rich text boxes and annotations currently inherit only part of their resolved typography. This is mostly pre-existing: `origin/master` applies family, size, color, and line height in Vue, but omits resolved font weight, font style, and stroke width. ByzHTML has the same limitation, with rich text export forcing normal weight/style and annotation export omitting resolved weight, style, stroke, and line height. This issue is more visible now because paragraph styles and direct rich-text-box overrides add more ways to set these values.This branch also introduces a new gap: direct rich-text-box style overrides need an export path. The fix is to resolve the active paragraph style font with `resolveFontStyle()` in rich text boxes and annotations, pass font weight, font style, and stroke width through CSS variables, and consume those variables in the editor CSS. ByzHTML can export resolved typography for rich text, annotations, and direct rich-text-box overrides, including resolved default rich-text font families through `cssFontFamily`.
3. Drop caps can now inherit a paragraph style with underline, but that underline does not apply consistently across all drop-cap surfaces. This is a new issue, since `origin/master` does not expose drop-cap underline through page setup or paragraph styles. The branch adds paragraph-style `textDecoration` and lets drop caps select paragraph styles, but drop-cap computed state still only tracks font, size, style, color, stroke, and line height. The Vue drop-cap component and exporters also have no text-decoration value to apply. The fix is to add `computedTextDecoration` and `computedTextDecorationPrevious` to `DropCapElement`, populate the value from the resolved drop-cap paragraph style in `LayoutService`, and include it in drop-cap layout change detection so paragraph-style underline changes trigger updates. The resolved text decoration can also be applied in Vue and emitted from ByzHTML and LaTeX when required. ByzHTML also can include the default drop-cap line height in the base drop-cap CSS.
4. Generated rich-text paragraph-style CSS and ByzHTML export do not consistently represent all resolved paragraph-style values. The missing `text-align` in generated rich-text paragraph-style CSS is a new issue because paragraph-style CSS generation is new in this branch. The raw `-webkit-text-stroke-width` values in ByzHTML are pre-existing: `origin/master` already emits several stroke widths without `px` units. This branch extends those export paths to paragraph styles and direct overrides, making the old unit serialization issue visible in more places. It also introduces a direct rich-text-box override export gap because rich text boxes gain direct font, color, stroke, and line-height override fields. The fix is to add `text-align: ${resolved.alignment}` to each generated rich-text paragraph-style CSS rule, centralize all ByzHTML stroke-width serialization through `getStrokeWidthCss()` so exported CSS includes `px` units, and export direct rich-text-box overrides as inline ByzHTML styles when they differ from the default rich-text paragraph style. Underline should stay out of generated rich-text paragraph-style class CSS because CKEditor handles underline as inline formatting, and explicit clears need to keep working.

danielgarthur:
Thanks. The style dropdown bug I fixed was a separate unrelated bug.

danielgarthur:
> commit 5445421f708809e377d5e8de189106ac2c4843f9 doesn't seem to work for me, as I still can't search for fonts. If I have time tomorrow I will try to debug.
> 
> 
> 
> I tried the audit I mentioned above and got these results:
> 
> 
> 
> 1. Some text formatting values can be changed on an element, but do not survive every element data path. Part of this is pre-existing: `origin/master` already has `lyricsTextDecoration` and text-box `lineHeight`, but lyric clipboard data does not include `lyricsTextDecoration`, and text-box clipboard data does not include `lineHeight`. This branch makes the problem worse: paragraph styles add `ParagraphStyleOverrides.textDecoration`, but element-level underline fields are not fully translated into that resolver shape. As a result, copied lyrics can lose underline state, including explicit underline clears, and copied text boxes can lose line-height overrides. Paragraph-style resolution can also miss explicit underline clears on lyrics or text boxes. The fix is to add `lyricsTextDecoration` to `NoteElement.getClipboardProperties()` and `lineHeight` to `TextBoxElement.getClipboardProperties()`, and to map lyric `lyricsTextDecoration` and text-box underline values into paragraph-style text-decoration overrides, using `null` to preserve explicit clears.
> 
> 2. Rich text boxes and annotations currently inherit only part of their resolved typography. This is mostly pre-existing: `origin/master` applies family, size, color, and line height in Vue, but omits resolved font weight, font style, and stroke width. ByzHTML has the same limitation, with rich text export forcing normal weight/style and annotation export omitting resolved weight, style, stroke, and line height. This issue is more visible now because paragraph styles and direct rich-text-box overrides add more ways to set these values.This branch also introduces a new gap: direct rich-text-box style overrides need an export path. The fix is to resolve the active paragraph style font with `resolveFontStyle()` in rich text boxes and annotations, pass font weight, font style, and stroke width through CSS variables, and consume those variables in the editor CSS. ByzHTML can export resolved typography for rich text, annotations, and direct rich-text-box overrides, including resolved default rich-text font families through `cssFontFamily`.
> 
> 3. Drop caps can now inherit a paragraph style with underline, but that underline does not apply consistently across all drop-cap surfaces. This is a new issue, since `origin/master` does not expose drop-cap underline through page setup or paragraph styles. The branch adds paragraph-style `textDecoration` and lets drop caps select paragraph styles, but drop-cap computed state still only tracks font, size, style, color, stroke, and line height. The Vue drop-cap component and exporters also have no text-decoration value to apply. The fix is to add `computedTextDecoration` and `computedTextDecorationPrevious` to `DropCapElement`, populate the value from the resolved drop-cap paragraph style in `LayoutService`, and include it in drop-cap layout change detection so paragraph-style underline changes trigger updates. The resolved text decoration can also be applied in Vue and emitted from ByzHTML and LaTeX when required. ByzHTML also can include the default drop-cap line height in the base drop-cap CSS.
> 
> 4. Generated rich-text paragraph-style CSS and ByzHTML export do not consistently represent all resolved paragraph-style values. The missing `text-align` in generated rich-text paragraph-style CSS is a new issue because paragraph-style CSS generation is new in this branch. The raw `-webkit-text-stroke-width` values in ByzHTML are pre-existing: `origin/master` already emits several stroke widths without `px` units. This branch extends those export paths to paragraph styles and direct overrides, making the old unit serialization issue visible in more places. It also introduces a direct rich-text-box override export gap because rich text boxes gain direct font, color, stroke, and line-height override fields. The fix is to add `text-align: ${resolved.alignment}` to each generated rich-text paragraph-style CSS rule, centralize all ByzHTML stroke-width serialization through `getStrokeWidthCss()` so exported CSS includes `px` units, and export direct rich-text-box overrides as inline ByzHTML styles when they differ from the default rich-text paragraph style. Underline should stay out of generated rich-text paragraph-style class CSS because CKEditor handles underline as inline formatting, and explicit clears need to keep working.

I skimmed this and I think most if not all are intentional or a misunderstanding. But I will review each carefully. 

It is intentional that drop caps and lyrics don't take alignment, for example. And drop caps don't allow underline. And lyrics don't consume line height because the line height is calculated so we can line up melismas. Line height would alter lyric vertical offset and there is already a separate setting for that. I think this is generally acceptable.

basil:
I looked at the changes to the file format and I wasn't sure why we are bumping it to 1.2. We haven't bumped the number recently, even though we have been making lots of file format changes, so I'm not sure what's different about this one.

The other question I had is why we're persisting local default style fields `color`, `fontSize`, `fontFamily`, `strokeWidth`, `fontSubfamily`, and `lineHeight` to rich text boxes. I asked AI to explain this to me but I'm not sure I understand the answer:

> The intended reason to persist those rich-text-box fields is stable “Default” formatting. In CKEditor HTML, unstyled text is represented by missing attributes, e.g. `<p>Plain text</p>`, so it needs an external fallback to know its family, size, color, and line height. Rich text boxes do not have a single object-level paragraphStyleId because paragraph styles live inside the HTML as neanes-style-* classes, so these fields would be local fallback overrides layered over DefaultText or Lyrics.
>
> These fields give you a useful capability: per-rich-text-box default styling without forcing inline spans onto every run of text. That is semantically clean if the product wants “this rich text box defaults to Alegreya 18pt” while still allowing paragraphs inside it to use classes and inline overrides. The downside is that the current diff looks incomplete: toolbar fallback appears to ignore these local overrides, root CSS does not fully apply fontStyle/strokeWidth, and export does not emit the per-box default styles. So this approach means committing to finish wiring.
>
> Removing them gives a simpler format: rich text styling comes only from paragraph classes in the HTML plus explicit CKEditor inline markup. That reduces schema surface, migration risk, and export ambiguity. The tradeoff is losing per-box default styling for unstyled rich text; changing DefaultText/Lyrics would affect inherited rich text unless the content has paragraph classes or inline spans. My recommendation: remove these persisted fields unless per-box rich-text defaults are an explicit feature for this PR; if they stay, finish the toolbar/root/export wiring before merging.

I didn't fully understand this but adding all these fields makes me nervous at a high level, and I wonder if the AI recommendation to remove these persisted fields is a good idea.

danielgarthur:
I didn't review the save service changes closely enough apparently. Because that doesn't make any sense to me either. 

For the 1.2 bump it was probably because AI knew this was a completely breaking change. In the past most changes I think were technically backwards compatible. But with these changes we are not even writing page setup style defaults anymore which is very much a breaking change. On the other hand, there should be no motivation for users not to use latest so it really doesn't matter.

basil:
I thought our general rule was "new versions must be able to read old content, but old versions might not be able to read new content." At least that's the rule I've been following recently. I didn't think we needed a version bump for this. I guess I assumed a version bump would be needed only for backwards-incompatible changes, but I'm not sure that's necessarily true either?

I asked Fable 5 for a second opinion about those new fields on rich text boxes, and it told me this:

> The paragraph-styles branch adds six persisted nullable style fields (`color`, `fontSize`, `fontFamily`, `strokeWidth`, `fontStyle`, `lineHeight`) to `RichTextBoxElement`. Their purpose is to be a per-box "default formatting" layer: unstyled CKEditor paragraphs are bare `<p>` tags with no attributes, so their appearance must come from outside the HTML — resolved from the built-in `DefaultText` style (or `Lyrics` for inline boxes) with these fields applied as a final override layer. Rich text boxes deliberately have no object-level `paragraphStyleId` (a test even pins this) because paragraph styles live inside the content as `neanes-style-*` classes.
> 
> The wiring is incomplete in three verified places. The main toolbar's "Default (…)" labels resolve the built-in style without the per-box overrides, while the properties pane and the rendered content include them, so the two UIs would disagree if a field were ever set. The container CSS applies only family, size, color, and line-height — never font-weight/style or text stroke — so `fontStyle` and `strokeWidth` do nothing on screen for unclassed text (a gap that partly predates this branch). And the ByzHtml exporter emits no per-box styles at all, even hardcoding `font-weight: 400; font-style: normal`, in contrast to the lyrics path which carefully emits per-element overrides. The strongest fact: **nothing ever writes these fields** — no UI, no migration, no code path — so they've been dead schema weight since the commit that introduced them, and the inconsistencies are only reachable via hand-edited save files.
> 
> Removing them is low-risk. They're unreleased (branch-only schema) and always null in practice, so deleting the fields and their save/load lines changes no observable behavior. Keeping them means committing to real scope with no driving feature: a UI to set them, fixing the toolbar fallback, adding root CSS for weight/style/stroke, and per-element style emission in the exporter. The only caveat is that removing them doesn't fix the pre-existing root-CSS gap for unclassed rich text — it just stops pretending per-box overrides feed into it.

basil:
I asked Fable 5 to look for other on-disk format bugs and it found these 3 additional issues. All 3 look legitimate to me.

## 1. Every new file writes stale `lyricsDefault*` / `dropCapDefault*` constants

The runtime `PageSetup` fields were deleted and `SavePageSetup` no longer assigns them — but the save-schema class (`src/models/save/v1/PageSetup.ts:69-75, 121-129`) still has live initializers (`'Source Serif'`, `fromPt(12)`, `fromPt(60)`, `'#000000'`, `0`), and the score is serialized with plain `JSON.stringify`. So every v1.2 file claims lyrics are Source Serif 12pt and drop caps 60pt as hardcoded constants, regardless of what the user's Lyrics/DropCap paragraph styles actually say. `textBoxDefault*` got the correct treatment — `undefined` initializers with a "deprecated load-only" comment — so this looks like a half-finished cleanup, presumably left partially intact because `dropCapDefaultLineSpan` is still genuinely used.

It's inert on round-trip through the new app (the constants happen to equal factory defaults, and file-saved styles take precedence on load), but it actively lies to old app versions and third-party readers, who read those fields as truth. The fix is one line per field (make them `undefined`); the loader already tolerates absence. Alternatively, back-filling them with the *resolved* style values would give old apps correct rendering — either is defensible, but stale constants are the worst of the three options.

## 2. Every note serializes `"lyricsParagraphStyleId": "lyrics"`

`SaveNote` writes it unconditionally (`SaveService.ts:1304`) even though the runtime default is `'lyrics'` and `LoadNote_v1` already handles absence (`SaveService.ts:2235`, `?? Lyrics`). Notes are by far the most numerous element — this adds ~35 bytes of noise per note against the format's own omit-when-default convention (master gated the whole lyrics-style block behind `if (!e.lyricsUseDefaultStyle)`; the very next line here is `spaceAfter: e.spaceAfter || undefined`).

One nuance: the unconditional write is load-bearing as written, because the legacy migration's guard is "field present → skip" (`migrateLegacyNoteLyricsParagraphStyleOverrides` returns early when `lyricsParagraphStyleId != null`). Omitting the default would still round-trip correctly — the migration reconstructs `'lyrics'` and preserves genuine overrides — but it would run migration logic on every load; the clean fix pairs omit-when-default with a version-based or score-level migration guard.

`SaveTextBox` (`SaveService.ts:1490`) and `SaveDropCap` (`SaveService.ts:1243`) have the same unconditional `paragraphStyleId` pattern, though those elements are few enough that it's a consistency nit rather than bloat.

## 3. Every rich text box serializes `"lineHeight": null`

`SaveRichTextBox` assigns it without `?? undefined` (`SaveService.ts:1560`), unlike `SaveTextBox` (`SaveService.ts:1511`). For paragraph *styles*, `lineHeight: null` is meaningful (explicit "normal" vs. inherit), but for rich text boxes null and absent are identical (`getParagraphStyleOverrides` maps null → undefined), so this is pure noise — and moot if the six orphan rich-text-box style fields are removed per my previous recommendation.

danielgarthur:
> I thought our general rule was "new versions must be able to read old content, but old versions might not be able to read new content."

That is the rule I have been trying to follow. I think you are correct we don't need the bump.

danielgarthur:
> I looked at the changes to the file format and I wasn't sure why we are bumping it to 1.2. We haven't bumped the number recently, even though we have been making lots of file format changes, so I'm not sure what's different about this one.
> 
> The other question I had is why we're persisting local default style fields `color`, `fontSize`, `fontFamily`, `strokeWidth`, `fontSubfamily`, and `lineHeight` to rich text boxes. I asked AI to explain this to me but I'm not sure I understand the answer:
> 
> > The intended reason to persist those rich-text-box fields is stable “Default” formatting. In CKEditor HTML, unstyled text is represented by missing attributes, e.g. `<p>Plain text</p>`, so it needs an external fallback to know its family, size, color, and line height. Rich text boxes do not have a single object-level paragraphStyleId because paragraph styles live inside the HTML as neanes-style-* classes, so these fields would be local fallback overrides layered over DefaultText or Lyrics.
> > These fields give you a useful capability: per-rich-text-box default styling without forcing inline spans onto every run of text. That is semantically clean if the product wants “this rich text box defaults to Alegreya 18pt” while still allowing paragraphs inside it to use classes and inline overrides. The downside is that the current diff looks incomplete: toolbar fallback appears to ignore these local overrides, root CSS does not fully apply fontStyle/strokeWidth, and export does not emit the per-box default styles. So this approach means committing to finish wiring.
> > Removing them gives a simpler format: rich text styling comes only from paragraph classes in the HTML plus explicit CKEditor inline markup. That reduces schema surface, migration risk, and export ambiguity. The tradeoff is losing per-box default styling for unstyled rich text; changing DefaultText/Lyrics would affect inherited rich text unless the content has paragraph classes or inline spans. My recommendation: remove these persisted fields unless per-box rich-text defaults are an explicit feature for this PR; if they stay, finish the toolbar/root/export wiring before merging.
> 
> I didn't fully understand this but adding all these fields makes me nervous at a high level, and I wonder if the AI recommendation to remove these persisted fields is a good idea.

Removed the rich text box fields in 090d7c1c66351eaa032cbd18cb0a0cede3abfac4.

danielgarthur:
Something I noticed: the editor seems to steal text-editing commands (like Ctrl + A, Ctrl + C, Ctrl + V) from the page setup dialog and styles dialog. This is more of a note for myself, but if you get to it before me, that's cool, too.

basil:
I reviewed this with Fable 5 and fixed most of the issues it brought up. It found a lot of interesting issues, ranging from subtle fallback issues to text box height recalculation bugs to duplicate style validation to copy/paste errors. I feel that the code is a lot more solid with a lot fewer rough edges after this pass.

Fable 5 did bring up a few legitimate issues that I did not tackle:

- `alignment` and `textDecoration` are dead properties for rich text boxes and annotations — no propagation when a style is edited. The generated per-style CSS deliberately omits `text-align`/`text-decoration`, and neither `TextBoxRich.vue` nor `TextAnnotation.vue` applies them. The only path applying a style's alignment/underline to rich text is one-shot materialization when the user picks the style from the dropdown — and underline there is add-only, never cleared when switching to a non-underlined style. So editing a style's alignment/underline updates plain text boxes but silently does nothing to rich text boxes, annotations, print/PDF, or ByzHtml export — while the Paragraph Styles dialog offers both controls for every style.
- Copy/paste of content that has a custom style from one score to another score that doesn't have that same style is problematic. I "fixed" this for most elements by downcasting to the closest built-in style, but I couldn't figure out what to do here for rich text boxes. Currently cross-score paste keeps unknown `neanes-style-*` classes in rich text HTML, which seems wrong. Probably need to implement the downast for rich text boxes as well. Alternatively, transfer the paragraph style into the new document for all object types.
- Exported `.annotation-container` CSS is dead code, should be cleaned up.
- `ParagraphStylesDialog` uses switches where the pending-until-update convention calls for checkboxes.
- In #2296 I went through some great pains to keep generated CKEditor HTML plain for bold and italic (documented in `advanced_typography.md`; this seems to be lost on this branch. Not essential but I think it's worth trying to keep the HTML clean in case we need to migrate away from CKEditor in the future.

danielgarthur:
Disclaimer: I put my response through AI to avoid you having to read my rambling and this is what it spit out. I am not aware of a good way to preserve the `<b>` unless maybe some CKEditor downcast/upcast magic.

The fundamental issue is that `<b>` only expresses **adding** bold. It cannot express **removing** bold that comes from the paragraph.

This works because the paragraph is normal by default:

```html
<p>Normal <b>bold</b> normal</p>
```

But if the paragraph style is bold:

```html
<p class="bold-paragraph">
  Bold by default <span>not bold</span>
</p>
```

there is no HTML equivalent of "not `<b>`". The child still inherits the paragraph's bold styling, so clearing it requires an explicit CSS reset (e.g. `font-weight: normal`).

The same limitation applies to `<i>` and `<u>`. They are one-way formatting elements: they can add formatting, but they cannot override formatting inherited from the paragraph.

basil:
Right, I understand that. My only point was that in the common case (where the parent style does not have bold or italic), it's possible to keep the markup simple, which I took some effort to do in #2296. Perhaps unnecessary, but a nice touch, and unfortunate (but acceptable) if it's lost in this PR. As a general programming best practice, I try to keep the common cases simple and reserve complexity for the edge cases, especially when it comes to serialization of user data. (See also Postel's law of "be conservative in what you send, be liberal in what you accept.")

danielgarthur:
I see. The thing about `<b>` is that it carries no explicit rendering info. Ultimately it comes from CSS _somewhere_. Either from the browser-level CSS or a CSS file in the project. Usually the browser default is `font-weight: bolder`, which doesn't even necessarily mean `bold` (as in `font-weight: 700`). Having the explicit span tells me "you WILL use font-weight 700", not simply "[bring attention to this element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/b)". I like the consistency that this brings (not that Neanes is cross-browser or that Chromium would ever change this default, probably). But explicit is good in my mind.

Of course, one could then make the argument that we should keep `<b>` and specify explicit styling for it in our app... So I don't know. Not having to mix `<b>` and `<span>` seems simpler, though, conceptually.

basil:
True in theory but in practice the distinction between e.g. "bolder" and "bold" is not an issue. I think keeping the default generated HTML simple and inline with how CKEditor normally emits things is a noble goal, though not a blocking one. This is somewhat subjective, but I think modern HTML markup is complex to the point of absurdity at times (just try reading **View Source** for google.com sometime...), and this complexity is often a source of interoperability issues, while simple markup is never a problem. But I digress, I've made my point that I think this is a desirable property to maintain, but not essential, and I trust your judgment.

danielgarthur:
> * `alignment` and `textDecoration` are dead properties for rich text boxes and annotations — no propagation when a style is edited. The generated per-style CSS deliberately omits `text-align`/`text-decoration`, and neither `TextBoxRich.vue` nor `TextAnnotation.vue` applies them. The only path applying a style's alignment/underline to rich text is one-shot materialization when the user picks the style from the dropdown — and underline there is add-only, never cleared when switching to a non-underlined style. So editing a style's alignment/underline updates plain text boxes but silently does nothing to rich text boxes, annotations, print/PDF, or ByzHtml export — while the Paragraph Styles dialog offers both controls for every style.

I fixed the alignment issue with with [cfd2401](https://github.com/neanes/neanes/pull/2400/commits/cfd2401dd6068a933bb1afe06026c4681d23ec8d). It's wild and makes me seriously consider why we don't just write our own text editor so we don't have to fight against CKEditor anymore.

As far as text decoration goes, my thought is this. If some user really decides to make a paragraph style with text decoration set to "underline". And that same user then decides to apply the style to a paragraph, and then decides that they want remove the underline from a single word in that paragraph... they're crazy. Moreover, there is a work around. The same effect could be achieved by leaving underline out of the paragraph style and underlining every word except the word they want to leave not underlined.

With that in mind, I'm going to clean up the underline mess I made early and have it function as I described above. That should address the "dead property" finding above.

basil:
Underline has no place in professional music typesetting, so I cannot concern myself with any issues regarding underline.

Nice job on the alignment plugin! It looks really great.

BTW I can reproduce my issue with searching for fonts in dialog boxes on `master`, so I don't think that problem is unique to this branch. :-( I'll try to debug it later.

basil:
> It's wild and makes me seriously consider why we don't just write our own text editor so we don't have to fight against CKEditor anymore.

https://prosemirror.net is just one AI-fueled bender away... :smile: I kid, I kid.

danielgarthur:
I attempted to restore `fontstyleediting.ts` to its former glory, but I gave up. There are extra considerations now with paragraph styles, and ultimately just using font-style invariably is simpler.

So with that, pending some more testing, I'm hoping this is done.

basil:
Any thoughts about the copy/paste issue?

danielgarthur:
Oh right. Also I didn't test what happens when you delete a style. I will have to think about it. My initial thought is that something should strip out unknown styles. I would think the CKEditor styles plugin would already do that. 

I don't think there should be an expectation that pasted text from another document carries paragraph style info.

basil:
I fixed several such bugs around deletion and paste with my Fable 5 improvement commits, but a few gaps remain, such as the one mentioned in my previous comment about rich text boxes.

danielgarthur:
For Copy Paste it depends on how easy it is. It would be nice if we could maintain the styling but I wouldn't consider it a blocker if it is too difficult. Maybe a custom paste handler that looks up the styling definition in the other open doc that it was copied from and copies missing styles?

If someone free copies from like a byzx file they are out of look though. But I think that is reasonable. 

I do think that the GUID class name should be dropped in favor of a sanitized class name based on style name. Since you made style names unique that should be fine now. Just need to prevent stupid stuff like blank spaces too. 

Then we first try to match on style name in the existing doc. If there is no match, then import from the other open doc. 

If the other doc was closed before paste, then maybe we can have a custom copy handler that saves the style info in memory?

basil:
I agree we shouldn't try to copy styles over. I actually like the downcast behavior better. The challenge is mapping styles appropriately, for example mapping a custom lyric style to the default Lyrics instead of just default text. Fable 5 wrote code to do most of this in my previous commits, but I didn't have the time/energy to do it for rich text as well.

danielgarthur:
> Currently cross-score paste keeps unknown neanes-style-* classes in rich text HTML, which seems wrong.

This does not seem to be the case anymore. Or if it is the case, I can't trigger it. Are you still able to get this to happen?

danielgarthur:
> I agree we shouldn't try to copy styles over. I actually like the downcast behavior better. The challenge is mapping styles appropriately, for example mapping a custom lyric style to the default Lyrics instead of just default text. Fable 5 wrote code to do most of this in my previous commits, but I didn't have the time/energy to do it for rich text as well.

By this do you mean if the style is a descendant of Lyrics, we set the paragraph style to Lyrics in the pasted text? That would be cool, I guess, but I don't think it's a blocker.

danielgarthur:
If that is what you mean, here may be a simple solution. Just encode the base style in the class name. E.g. `neanes-style-base--lyrics--name--my-custom-lyric-style`. There are some details to hammer out, but this is a start. Could also maybe use multiple classes to encode everything, but that might require a custom styles plugin for ckeditor.

basil:
> This does not seem to be the case anymore. Or if it is the case, I can't trigger it. Are you still able to get this to happen?

Yes, I can still reproduce this by creating a custom style, then pasting it into a new document and saving. It's important not to click into the new text box before saving. AI analysis below:

**Problem**

The paste flow can leave dangling paragraph style references inside rich text content. When a `RichTextBox` or note annotation contains a class like `neanes-style-<uuidA>`, that UUID can be pasted into another document that does not define the style. The document then saves a BYZX file with a reference to a nonexistent style. This does not crash the editor, but it violates the no-dangling-references invariant and causes weaker fallback behavior than other text surfaces.

**Cause**

Paste-time sanitization is the only point where clipboard style references are normalized before entering the target document. The sanitizer handles `Note`, `TextBox`, `DropCap`, and `AlternateLine`, but it does not handle `RichTextBox`. The `Note` path also misses `note.annotations`, including annotation text on alternate-line notes. Because rich text content and annotations are cloned and saved verbatim, any unsanitized `neanes-style-*` class persists into the model and saved file.

**Suggested Solution**

Add a `RichTextBox` case to `sanitizeElementParagraphStyleIds` so paragraph style classes inside rich text HTML are rewritten during paste. Also sanitize `note.annotations[].text` anywhere note content is handled, including alternate-line notes. The shared rewrite logic should replace any `neanes-style-*` class whose ID is missing from the target style set with the appropriate surface fallback class. Generalizing `rewriteRichTextHtmlForDeletedStyles` would keep deleted-style cleanup and cross-document paste cleanup consistent.

basil:
In addition to that, Fable 5 found (against the latest code) and I confirmed (with manual testing) this other bug:

# Bug: Rich text Bold/Italic toggles ignore the paragraph style's font style

## Summary

The Bold (B) and Italic (I) toggle buttons in the rich text toolbar (and
their Ctrl+B / Ctrl+I shortcuts) compute their pressed state and toggle
target from the selection's explicit inline `fontStyle` attribute only,
falling back to Regular. They never consult the resolved paragraph style.
The font-style select in the same toolbar does fall back to the paragraph
style, so the two widgets disagree, and un-bolding paragraph-styled bold
text takes two clicks.

## Steps to reproduce

Setup -- make any style bold (no built-in style ships bold; Title is just
centered 28pt):

1. Open Format > Paragraph Styles... and select "Title" (or click
   "New Style").
2. Set its font style to Bold and close the dialog.

### Repro A: toggle state is wrong

3. Insert > Rich Text Box, type a few words, and click into the box so the
   rich text toolbar appears.
4. Apply "Title" from the paragraph-style dropdown in the toolbar. The text
   renders bold, and the font-style select next to the family picker
   correctly reads "Bold".
5. Look at the B toggle button in the same toolbar.

Expected: B is pressed, consistent with the font-style select.

Actual: B is not pressed, even though the text is effectively bold.

### Repro B: un-bolding takes two clicks

6. Select all the text in the box (drag or Ctrl+A inside the editor; a
   collapsed caret would only affect newly typed text).
7. Click B once.

Expected: the text un-bolds.

Actual: nothing visible changes. The text was already bold from the style's
class CSS; this click adds an explicit inline `fontStyle=Bold` on top of it
(the command computes its toggle target from Regular, so Regular -> Bold).
The only feedback is the button lighting up.

8. Click B again. Now the text actually un-bolds: the explicit Bold toggles
   to an explicit Regular, which pins `font-weight: 400` and overrides the
   class CSS.

## Root cause

- The toggle button's pressed state comes from `FontStyleToggleCommand.value`,
  computed from `effectiveFontStyle()`
  (src/ckeditor-plugins/fontstyle/fontstylecommand.ts:159).
- `effectiveFontStyle()` falls back to `DEFAULT_FONT_STYLE`
  (src/ckeditor-plugins/fontstyle/fontstylecommand.ts:56) and never consults
  the resolved paragraph style.
- The font-style select in the same toolbar falls back to
  `resolvedActiveParagraphStyle.fontStyle`
  (src/composables/useRichTextStyleCommands.ts:177-183), which is why the
  two widgets disagree. The same fallback chain needs to reach the toggle
  command.
  
## Notes

- Ctrl+B reproduces the same behavior; it executes the same command
  (src/ckeditor-plugins/fontstyle/fontstyle.ts:24).
- The identical failure exists for I / Ctrl+I with a style whose font style
  is Italic.
- The toggle is only enabled when the effective font family has a Bold face
  (src/ckeditor-plugins/fontstyle/fontstylecommand.ts:152-157), so leave the
  default font family in place when reproducing -- it has one.
- Applies to both element types that show the paragraph-style select: rich
  text boxes and annotations (src/components/RichTextToolbar.vue:468-472).

basil:
I haven't checked this empirically but it seems plausible, I read the code and saw that the existing code is a string matching heuristic, so makes sense that it would no longer match with the addition of `neanes-style-header`:

> We've lost automatic odd/even header setup for rich-text headers, which is a regression from `origin/master`: existing score files may contain app-generated rich-text headers and footers that users never edited, and in `origin/master`, when a user enabled facing pages, the app recognized those untouched defaults and automatically turned on separate odd- and even-page headers so page numbers and running text appeared correctly on left and right pages; the new branch changes the HTML generated for default headers and footers by adding paragraph-style CSS classes such as `neanes-style-header`, but the code that decides whether a header or footer is still the default still relies on an exact HTML string match, so legacy files containing the older default HTML are now mistaken for user-customized content, causing the app to skip the automatic odd/even setup and lose the expected facing-page header behavior for those older files.

danielgarthur:
> I haven't checked this empirically but it seems plausible, I read the code and saw that the existing code is a string matching heuristic, so makes sense that it would no longer match with the addition of `neanes-style-header`:
> 
> > We've lost automatic odd/even header setup for rich-text headers, which is a regression from `origin/master`: existing score files may contain app-generated rich-text headers and footers that users never edited, and in `origin/master`, when a user enabled facing pages, the app recognized those untouched defaults and automatically turned on separate odd- and even-page headers so page numbers and running text appeared correctly on left and right pages; the new branch changes the HTML generated for default headers and footers by adding paragraph-style CSS classes such as `neanes-style-header`, but the code that decides whether a header or footer is still the default still relies on an exact HTML string match, so legacy files containing the older default HTML are now mistaken for user-customized content, causing the app to skip the automatic odd/even setup and lose the expected facing-page header behavior for those older files.

I don't think this is worth fixing. It only affects users if they open an old document and switch that document to facing pages. And even then, the user can still check "Use different odd/even pages" manually. 

If anything, I'm not sure I really like the whole "auto-enable different odd/even" because it is a little too magical and doesn't tell the user what's happening, especially since "facing pages" and "odd/even" are in different sections of the dialog. Moreover, if I choose "facing pages" in a new document and then uncheck "odd/even", then save, it re-checks it, which is a bug. But these are all edge cases.

basil:
> I don't think this is worth fixing. It only affects users if they open an old document and switch that document to facing pages. And even then, the user can still check "Use different odd/even pages" manually.

Makes sense. In that case, if we don't want to support the feature, then I think this PR should delete the heuristic rather than introducing a code change that breaks a previously-working heuristic.

---

The very first commit in this PR replaced the strict CKEditor-Reka contract with a looser last-active contract. This went against the model carefully documented in `notes/ckeditor_reka.md`. I only noticed this when I saw Fable 5 later refactoring everything into the new last-active contract. Looking into it, this was a workaround apparently made in the initial commit for a focus problem where there was a much cleaner solution. Implemented the cleaner solution in commit 92d6251f and restored the original strict semantics to prevent regressing the clean design introduced in #2279.

---

Did another pass with Fable 5, I've used up most of my Fable 5 credits for the week but I think the code is much cleaner now. It's hundreds of lines less code, too.

--- 

It's strange to me that the switches disappear completely from the Format Paragraph box when there is no parent, but then re-appear once you change an attribute like Font Size. We should pick a model here: either parentless styles are fully defined without individual components being togglable, or they behave like everything else with individual styles being togglable, but the current UX is very much a halfway state that is inconsistent.

---

The switches in the UI seem to go against the general paradigm of using checkboxes for temporary uncommitted state and switches for state that is committed immediately without confirmation.

---

This PR introduces a number of tests that use mocks/spies/stubs with using Vitest’s `vi.mock`, `vi.fn`, `vi.hoisted`, and `vi.spyOn(...).mockReturnValue`. We're even pulling in a new dependency, `jsdom`, for testing purposes. Not sure what your opinion is about this type of test as a programmer, but I feel very strongly against including such tests. They have very low long-term value, rarely catch regressions, and almost always break during unrelated refactorings. When I saw them in Jenkins-land, I would delete them on sight. Worse, they can mislead AIs into thinking that some internal behavior is an API contract of some sort that should be "preserved" when this is not the case. I would recommend deleting such tests entirely. My opinion about software testing is that pure unit tests are good, but integration tests should be done with something like #2382 rather than writing fragile mocks that almost always have to be rewritten or deleted later. That kind of integration test usually _does_ catch real issues, unlike mock-based tests which are very low signal.

danielgarthur:
Thanks for all the cleanup.

I don't have a good feeling yet for when I should just "trust" unit tests that AI creates vs reviewing all of them. Probably I should error on the side of reviewing all of them, or create a skill to prune dumb tests (or prevent them from being written).

> The switches in the UI seem to go against the general paradigm of using checkboxes for temporary uncommitted state and switches for state that is committed immediately without confirmation.

I don't have a strong feeling either way, but I was not aware of the paradigm. I just thought it was a subjective concept of "switches make things immediately happen". So toggling the switch makes the field immediately editable. As opposed to, say, toggling header settings in Page Setup, which sets options that take effect later. Although I could see another point of view which is that the override itself doesn't take effect immediately within the document. If you feel really strongly, I will change it.

danielgarthur:
I have no idea what the deal was with `package-lock.json`. I kept running `npm i` and seeing no differences locally, but the CI build was failing. Eventually I determined that `nvm` was overriding `corepack`. I switched to npm v11.12.1 and ran `npm i` and got all those changes. But now after pulling down your commit, `npm i` has no changes.

basil (dismissed):
Awesome work, this will be a great feature! And thanks for your patience as I reviewed this.

LGTM; might be worth doing one more round of manual testing after all the recent AI cleanups.

danielgarthur:
Thanks for all the polishing and feedback. 

There is still one thing that I'm questioning now. Originally, I had `Default Text` as the only parentless style. Then at some point I made it so that everything can be parentless. Now I am questioning that decision. I thought I had a motive, but I have since forgotten it. It seems like a needless complication since inheriting from `None` is equivalent to inheriting from a hard-coded `12pt Source Serif` font. It seems like users could just as easily inherit from `Default Text` with no loss in available functionality. Thoughts?

I've given some thought about copy/paste. I will probably make a branch to test out my idea, but ultimately it will be part of a second PR. My idea is that whenever the user copies, `TheEditor` can look at the selected elements, find all relevant styles (including the hierarchies) and put them on its `clipboard` along with the elements. Then on paste, we can match on style name in the target document, and if no name matches, import the missing styles.

This would not handle the edge case of when a user copies text out of a rich text box, then closes the app. Then reopens and pastes. In this case, there would be nothing in the Editor's clipboard. I think it would still be possible to handle this case by intercepting the copy action or using CKEditor's Clipboard pipeline, then adding extra metadata to the operating system clipboard. I don't think I will bother with that scenario first pass, though.

basil:
> It seems like a needless complication since inheriting from `None` is equivalent to inheriting from a hard-coded `12pt Source Serif` font. It seems like users could just as easily inherit from `Default Text` with no loss in available functionality. Thoughts?

Agreed, that does sound simpler and more intuitive. The hard-coded 12pt Source Serif is effectively a "root below the root" which is not very intuitive. Seems simpler to have an explicit single root in all cases.

PR #2416, opened 2026-07-02
fix: keyboard shortcuts in dialogs

danielgarthur:
## Summary

This PR fixes edit keyboard shortcuts while dialogs are open, so standard actions like copy, cut, paste, select all, undo, and redo work more consistently in dialog text fields.

## Motivation

The editor’s global keyboard handling was designed around the score surface, which meant dialog-focused input could either miss expected edit shortcuts or risk falling through to score/editor shortcuts. This change gives dialogs a small, explicit edit-shortcut path and then stops further editor shortcut processing while a dialog is open.

## Major Changes

- Adds dialog-aware handling for standard edit shortcuts.
- Refactors Mac text-input shortcut handling into a shared helper.
- Leaves browser-native paste alone for normal dialog inputs, while still routing Electron/contenteditable paste through the app paste path.
- Avoids treating non-Mac `Ctrl+Alt` combinations as platform edit shortcuts.

## Implementation Notes

- `handleEditShortcut()` now returns whether it consumed the shortcut, replacing the old local `handled` flag.
- Dialog handling only calls `preventDefault()` when the app actually handles the shortcut.
- Platform shortcut detection is centralized so Mac uses `Meta`, while non-Mac uses `Ctrl` without `Alt`.

## Reviewer Notes

- Dialog input behavior in Electron and browser builds.
- Mac vs Windows/Linux shortcut behavior.
- Paste behavior in normal inputs vs contenteditable fields.
- Any unintended interaction with native text-editing shortcuts or international keyboard layouts.

## Testing

- `git diff --check` passed during review.
- No automated test suite was run.
- Manual testing in MacOS and Linux.


basil (approved):
I found a small bug fix (tested manually, reproduced before the fix and verified fix works) as well as a small cleanup with Fable 5; implemented both. Feel free to revert if you don't like either.

PR #2418, opened 2026-07-02
feat: Make neume selector layout fluid

### LINKED ISSUE #2250: New column with "descending neumes with petasti" in left panel
Thank you for your awesome software. It is far the better solution for byzantine notation.

**Is your feature request related to a problem? Please describe.**
Sometimes I need to enter "descending neumes with petasti" which are located in the second column of the left panel, below "ascending neumes with petasti". This requires scrolling up & down on the left panel, especially in my 13" laptop.

**Describe the solution you'd like**
I would suggest "descending neumes with petatsti" to be moved in a forth column after "descending neumes" to avoid much scrolling. In addition the size of neumes can be reduced a bit to shorten left panel width.

**Describe alternatives you've considered**
None

**Additional context**
 I created an image with my desired layout which I attach here.
Image

basil:
Fixes #2250

The quantitative neume selector was a fixed four-column layout: when the side panel was narrow, part of the palette was hidden behind a horizontal scrollbar, and when it was wide the extra space went unused. This makes the palette reflow to fit the panel:

- Lay the palette out as a CSS grid whose column count is recomputed from the panel width via a `ResizeObserver`, and again whenever the default neume font family changes.
- Size columns from the glyph advance widths of the selected neume font (falling back to the bundled Neanes metrics), so wide glyphs get the room they need.
- Preserve the semantic grouping while reflowing: the palette is defined as five source groups (the petasti group is split into ascending and descending halves). When fewer columns fit, adjacent groups are merged smallest-pair-first; when more columns fit, the largest columns are split in half.
- Open the vareia-dotted and kentemata combination submenus downward when there is not enough room above the trigger's row, reversing the item order in that case so the default item stays under the cursor, generalizing the previously hardcoded up/down menu directions.
- Note to reviewers: Suggest reviewing with **Hide Whitespace** enabled.

### DIFF
--- a/src/components/NeumeSelector.vue
+++ b/src/components/NeumeSelector.vue
@@ -594,7 +594,7 @@ const showVareiaDottedMenu = ref(false);
 const selectedSecondaryGorgon = ref<SecondaryGorgonMenuItem | null>(null);
 const selectedVareiaDotted = ref<QuantitativeNeume | null>(null);
 const panelElement = ref<HTMLElement | null>(null);
-const columnCount = ref(5);
+const currentColumnCount = ref(5);
 
 const sourceMatrixColumns: QuantitativeNeume[][] = [
   ascendingNeumes,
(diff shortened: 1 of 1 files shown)


danielgarthur:
Not sure what this means: 

> Note to reviewers: Suggest reviewing with Hide Whitespace enabled.

danielgarthur (dismissed):

basil:
That just means that some of the change reindents a lot of content, which makes the diff hard to read. It can be viewed more easily on the command line with `git show -b` (`--ignore-space-change`), or in the GitHub UI by checking **Hide Whitespace** in the settings tab of the GitHub PR UI.

After using this for a day or so, I noticed that I had accidentally shifted the positions of a few of the neumes in the default view, so back to Draft until I can fix that.

PR #2428, opened 2026-07-05
fix: font styles for rich text neumes

### LINKED ISSUE #2425: Rich text box plagal style bug
**Rich text box plagal style is bold before tab change (v0.5.39):**

[bug_rich_text_box_plagal_style_v0.5.39.byzx] - [source](https://github.com/ilizol/neanes-examples/blob/42196f0acfb8ab86a7a0e989a96753eb3803067e/bugs/bug_rich_text_box_plagal_style_v0.5.39.byzx)
[bug_rich_text_box_plagal_style_v0.5.39_before_tab_change.png] - [source](https://github.com/ilizol/neanes-examples/blob/42196f0acfb8ab86a7a0e989a96753eb3803067e/bugs/bug_rich_text_box_plagal_style_v0.5.39_before_tab_change.png)

---

**Rich text box plagal bold style is cleared after tab change (v0.5.39):**

[bug_rich_text_box_plagal_style_v0.5.39_after_tab_change.png] - [source](https://github.com/ilizol/neanes-examples/blob/42196f0acfb8ab86a7a0e989a96753eb3803067e/bugs/bug_rich_text_box_plagal_style_v0.5.39_after_tab_change.png)

danielgarthur:
Closes #2425

### DIFF
--- a/src/ckeditor-plugins/fontstyle/fontstyleediting.ts
+++ b/src/ckeditor-plugins/fontstyle/fontstyleediting.ts
@@ -41,8 +41,9 @@ import {
 const FONT_FAMILY = 'fontFamily';
 
 // Transient markers used only during upcast to carry a legacy <strong>/<em> (or
-// a bundled-style font-weight/font-style) onto the wrapped text. The post-fixer
-// folds them into fontStyle and removes them, so they never survive a save.
+// a bundled-style font-weight/font-style) onto text and inline neume objects.
+// The post-fixer folds them into fontStyle and removes them, so they never
+// survive a save.
 const LEGACY_BOLD = 'neanesLegacyBold';
 const LEGACY_ITALIC = 'neanesLegacyItalic';
 const LEGACY_NORMAL = 'neanesLegacyNormal';
@@ -566,7 +567,7 @@ export default class FontStyleEditing extends Plugin {
     let desired: string | null;
 
     if (fontStyle == null) {
-      // Fresh from upcast (including neumes) or default-font text: derive the
+      // Fresh from upcast, including neumes, or default-font text: derive the
       // font style from the face name, then fold in any legacy markers.
       if (family == null || family === '') {
         newFamily = null;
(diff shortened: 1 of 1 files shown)


danielgarthur:
@basil This change removes some code with a comment that makes me think something intentional was going on with the neumes, but I'm not sure what or whether I broke something. It seems like neumes were intentionally excluded from the font style plugin, but I don't know why.

basil:
Sorry, this doesn't ring a bell. :-(

danielgarthur:
As far as I can tell it's just an oversight. But if you recall anything let me know. 

GPT 5.5 had this to say. 

> After reading `notes/advanced_typography.md`, I’d call the old behavior an implementation oversight, not a product-intent exception.
> 
> The old code was intentional in a narrow engineering sense: `FontStyleEditing` was originally written around CKEditor text runs. Its comments say legacy markers carry `<strong>/<em>` or CSS weight/style “onto the wrapped text,” and the old branch deliberately treated non-text items as cleanup-only. So the removed comments were not accidental noise; they reflected the implementation’s old mental model.
> 
> But the typography notes point the other way for actual semantics:
> 
> - Rich text normalizes legacy `<b>`, `<em>`, `font-weight`, and `font-style` into `fontStyle` on import.
> - Explicit `fontStyle` must serialize as CSS that pins both axes.
> - Legacy markers are import syntax only and should be folded into canonical `fontStyle`.
> - `insertneumeediting.ts` is explicitly listed in the notes as allowing `fontStyle` on inline neume objects.
> 
> There’s no documented carve-out saying “except inline neume widgets.” The old code looks like it noticed neumes might appear in the normalization range, but only stripped markers from them rather than finishing the same fold-to-`fontStyle` process. That explains your reload bug exactly: the neume was allowed to have `fontStyle`, and live editing could set it, but legacy CSS import/downcast normalization treated it as a non-text object and dropped the compatibility signal.
> 
> So my read is: intentionally defensive old implementation, semantically an oversight once inserted neumes became rich-text inline objects that can carry `fontStyle`.

PR #2440, opened 2026-07-07
feat: paragraph style paste

danielgarthur:
## Summary

This PR preserves custom paragraph styles when copying, cutting, pasting, and copying/pasting formats between scores. Instead of falling back to built-in paragraph styles whenever pasted content references unknown style IDs, the editor now carries the relevant custom style definitions through the internal clipboard and imports or remaps them into the destination score.

## Motivation

Paragraph style IDs are score-local. Before this change, pasting content that referenced a custom paragraph style into another score could lose that styling because the destination score did not know the source style ID. This was especially visible for text boxes, drop caps, note lyrics, annotations, rich text boxes, and copied formats that depended on custom paragraph styles.

## Major Changes

- The internal clipboard now stores both copied score elements and the custom paragraph styles those elements reference.
- Paste operations resolve copied paragraph styles against the destination score, reusing an existing custom style by matching display name when possible and importing missing custom styles otherwise.
- Paragraph-style references in pasted elements and rich text classes are rewritten through the resolved ID map before insertion or replacement.
- Copy Format / Paste Format now carries the paragraph style definition needed by copied text box or note formatting.
- Paste without lyrics uses the same `includeLyrics` clone behavior for style collection as it uses for the pasted elements, so lyric-only paragraph styles are not imported when lyrics are intentionally omitted.
- Paste operations that import paragraph styles batch the score-style update with the element/format update and trigger the existing resizable text recalculation path.

## Implementation Notes

- `clipboardParagraphStyles.ts` now owns three related steps: collecting referenced custom styles and ancestors, resolving them against the destination score, and rewriting pasted style references.
- Built-in paragraph styles are not imported; only custom styles are carried through the clipboard.
- Custom style matching is by trimmed display name. If the destination already has a same-named custom style, the pasted content remaps to that style instead of creating a duplicate.
- Imported parent chains are preserved where possible. Missing or cyclic parents fall back to `Default Text`.

## Known Limitations

- This preserves styles for the app’s internal clipboard state. It does not add OS clipboard serialization for paragraph style definitions across app instances.


basil (commented):
Looks good overall, I haven't tested this manually yet. A quick run of Opus 4.8 found a handful of issues, these seem plausible, and I think should likely be fixed, especially if they leave behind dirty data in saved documents.

---

The staged clipboard paragraph-style import code in `src/utils/clipboardParagraphStyles.ts` can create an unused imported ancestor style. `collectCustomParagraphStylesFromIds` emits ancestor styles before descendant styles, so a copied custom `Leaf` style with a custom `Base` parent is resolved as `Base` first and `Leaf` second. If the destination score already has a custom style named `Leaf` but does not have the copied `Base`, `resolveClipboardParagraphStyles` imports `Base` before it discovers that `Leaf` should be reused by name. When `Leaf` is remapped to the existing destination style, no pasted element and no imported descendant references the freshly imported `Base`. The result is persistent style-list pollution and a dirty document even though the extra style has no visible effect on pasted content. The cause is that the top-level resolver commits each import decision in collection order instead of resolving from the actually referenced descendant to the ancestors it still needs. The fix is to route every unresolved clipboard style through the recursive resolver, or otherwise defer parent import decisions until the descendant's final target is known. When a source style is reused by target name, its source-only ancestor chain should be skipped unless another imported style still references that ancestor.

`onFileMenuPasteFormat` in `src/components/TheEditor.vue` can import paragraph styles even when no selected element receives the copied format. The function resolves the copied format's paragraph styles and pushes an `update-properties` command before it checks whether the current selection contains any element of the copied format type. If a text-box format with a custom style is pasted onto a range containing only notes, `applyCopiedFormat` never adds a text-box update, but the paragraph-style import command remains in the command list. The same problem occurs when the user has a single selected note and pastes a copied text-box format, or the reverse pairing of note format and selected text box. The subsequent `commands.length` check executes the import-only command and then calls `save()`. The user sees no formatting change, but the destination score gains copied custom styles and becomes dirty. The cause is that style import commands and target update commands share one array without tracking whether any target update was produced. The fix is to collect or count `applyCopiedFormat` commands first, then prepend the paragraph-style import only if at least one element will actually be updated.

The resolver contract is unclear because current callers pass only collected custom styles, while `resolveClipboardParagraphStyles` also contains branches for broader input shapes. The clipboard collection helpers filter out built-in paragraph styles before calling the resolver, so the exported resolver's `builtIn` branch is not reached by the current copy, paste, or copy-format paths. The exported resolver's same-id target branch is still reachable because copying and pasting within the same score supplies custom clipboard style ids that already exist in the target score. Inside the recursive helper, same-id handling is mostly defensive in the current implementation because parent resolution checks target ids before it recurses into the helper. This mixture makes the resolver harder to audit because some branches describe current behavior and others describe defensive behavior for inputs current callers do not provide. The cause is that the public resolver accepts any `ParagraphStyle` array while its actual callers pass a narrower collected-custom-style set, and the top-level loop duplicates logic from the helper. The fix is to either narrow the helper contract and remove unreachable built-in handling, or keep the broader contract and document which branches are intentionally defensive. Reusing the recursive helper from the top-level loop would also put target-id reuse and name reuse in one place instead of spreading them across duplicated conditionals.

basil (approved):
Manually tested, LGTM

I pushed a few refactoring commits, feel free to revert if you don't like any

Issue #2387, opened 2026-06-28
Save combinations


ichnilatis-gr:
I would like to ask if it is possible to save a combination of neumes, especially after fine-tunings of their positions, so that we can always use them with the same distances.
I think it would be very useful and convenient if we had the possibility to add our combinations to the "Common Combinations" toolbar.
Thank you.

_Originally posted by @ichnilatis-gr in https://github.com/neanes/neanes/discussions/2290_

danielgarthur:
Duplicate of #1836

Issue #2305, opened 2026-06-17
Set options for sharp position with Double Apostrophos


ichnilatis-gr:
When we add a sharp to Double Apostrophos, the sharp is placed under the first Apostrophos. Would it be possible to have the option to choose which Apostrophos the sharp should be added to, as is the case with other combinations?

<img width="130" height="91" alt="Image" src="https://github.com/user-attachments/assets/a945a370-cc13-4602-b53e-e13e0ea069ee" />

<img width="141" height="93" alt="Image" src="https://github.com/user-attachments/assets/496f606c-10f3-47d6-b54e-89b33aa3ab63" />

Thank you!

Issue #1865, opened 2025-12-05
Sharp with apostrophos


gugushvili:
A sharp in this combo of apostrophoi must go by default with the second one (also in the playback).

<img width="88" height="51" alt="Image" src="https://github.com/user-attachments/assets/e0620299-88dd-42a7-86f9-8ff3cfef01c2" />

PR #2447, opened 2026-07-07
feat: user combos

### LINKED ISSUE #1836: Common combos
Now that the program supports common combos, is there a way to introduce user-defined common combos (and maybe also to associate them with keyboard shortcuts)? That would seem like a useful feature.

danielgarthur:
Closes #1836

## Summary

Adds user-defined neume combinations. Users can save a selected multi-note passage as a reusable combo from the score context menu, see saved combos alongside the built-in combo presets, insert them through the existing combo selector, and delete saved combos from that selector.

## Motivation

The built-in neume combinations are useful, but they only cover a fixed set of common patterns. This change lets users build a personal library of repeated neume patterns without expanding the app’s saved score format or requiring new built-in presets for every workflow.

## Major Changes

- Adds a “Save as combo” context-menu action for selections containing at least two note elements.
- Adds `useNeumeCombinations`, which combines built-in combos with user combos persisted in `localStorage`.
- Updates the combo selector to render both built-in and user combos, with a visual marker and delete control for user-created entries.
- Adds a confirmation dialog before deleting a user combo.
- Extends neume-combination model types with stable ids, source metadata, and a serialized note payload for local persistence.
- Adds localized strings for the new save/delete combo UI.

## Implementation Notes

- User combos are stored outside the document in `localStorage`, so they behave like editor preferences/personal presets rather than score data.
- Saved combo notes are serialized through a dedicated neume-combination payload instead of storing full score elements. This keeps the feature focused on reusable neume notation data rather than lyrics or broader document structure.
- Insertion reuses the existing paste path by temporarily putting the combo elements into the clipboard state, so combo insertion follows the same insert/edit/auto behavior as normal pasted elements.

## Testing

- Manual testing


basil:
This is cool I suppose, but I can't see myself using it, in part because I obsessively run `rm -rf ~/.config/Neanes` and would therefore lose any combinations I saved. The lack of any search functionality means it's limited to a handful of combinations anyway, and it's easier for me to just type them out.

What I _would_ actually use is something like https://sourceforge.net/projects/psaltichelper that could load and save libraries of formulas, with accentuation patterns and other metadata, a search interface, etc. But that is a far more ambitious project.

image

ichnilatis-gr:
Thank you for implementing this feature. I am looking forward to using it in the next version. 

In response to @basil, I would like to explain [that my request](https://github.com/neanes/neanes/issues/2387) for this feature was prompted by the fact that we currently cannot fine-tune the position of certain combinations of neumes and modifiers without using `Positioning`, like the ones mentioned in #2305 and #1865. With this feature, however, we will be able to save our customized combinations of neumes and modifiers after adjusting their positions with the `Positioning` feature, and then reuse them in other scores.

Thank you.

basil:
Motivation should be provided in the initial ticket, not after the fact in a comment on a closed PR. In any case, this does not seem like a legitimate use case to me. Adding a feature to compensate for a bug is bad for long-term maintenance, as once users start relying on the feature it cannot actually be removed even after the bug is fixed.

danielgarthur:
Indeed. Although I never intended this to be used in that way. It was always my plan to allow users to add combos when I added the combo feature, but not to compensate for bugs or positioning issues...

ichnilatis-gr:
I apologize. However, I was not the only one who asked for this feature. Moreover, I believe that the more I use the app, the more I'll find this feature useful in other situations as well.
In any case, I'm sorry if I caused you any trouble.

danielgarthur:
No problem. Now I know what bug I need to go fix 👍

PR #2448, opened 2026-07-08
feat: outline color

### LINKED ISSUE #2392: Custom Outline Color for Drop Caps
Sorry for bothering you again, but may I make another request?

Regarding the letter styling and especially the styling of the drop caps, would it be possible to have an outline in a different color? For example, the letter could be white and the outline red.

Thank you!

danielgarthur:
Closes #2392.

## Summary

Adds outline color support for paragraph-style driven text, including lyrics, text boxes, drop caps, and rich text paragraph styles.

## Motivation

Outline width could already be controlled, but the outline color was effectively tied to browser defaults or text color behavior. This change makes outline color an explicit part of the paragraph style model while preserving “Same as text” as the default behavior.

## Major Changes

- Adds `strokeColor` to paragraph style overrides, element-level overrides, and the v1 save/load schema for lyrics, text boxes, and drop caps.
- Adds outline color controls to the Paragraph Styles dialog and the Lyrics, Text Box, and Drop Cap properties panes.
- Introduces a dedicated `StrokeColorPicker` with a “Same as text” option backed by `currentcolor`.
- Threads resolved outline color into editor rendering, layout change detection, rich text paragraph CSS, and ByzHTML export.
- Adds save/load tests for paragraph-style and element-level stroke color round trips.

## Implementation Notes

- `currentcolor` is used as the fallback/default stroke color so existing behavior remains “outline follows text color” unless the user explicitly chooses a custom color.
- Outline color follows the existing paragraph-style override pattern, including clear-formatting behavior and per-element overrides.
- The new UI is intentionally placed in the dialog/properties panes alongside outline width, rather than adding toolbar controls.

## Testing

- Manual testing

## Known Limitations

- Rich text boxes cannot override the outline color. This limitation already existed for outline width. The workaround is to create separate styles for different outline properties.


src/services/integration/ByzHtmlExporter.ts
```diff
@@ -357,13 +357,16 @@ export class ByzHtmlExporter {
         color: ${lyricsStyle.color};
         font-weight: ${defaultLyricsFont.cssFontWeight};
         font-style: ${defaultLyricsFont.cssFontStyle};
+        -webkit-text-stroke-width: ${lyricsStyle.strokeWidth};
```
danielgarthur, line 360: Stroke width seems to have been missing before, which was a prior bug fixed by this addition.

PR #2495, opened 2026-07-18
feat: Romanian and Slavonic lyrics

### LINKED ISSUE #712: Unexpected result with Greek lyric melisma
Many thanks for the Greek melisma support. It helps a lot, but cases like https://github.com/basil/byzantine-music/blob/485268d11cca0e5a65377254782680d78313b58a/Mysteries%20and%20Sacraments/Paraklesis/36%20Speechless%20be%20the%20lips.byzx still pose a challenge—entering "Eις πολ--λα________ ε-----τη___ Δε------σπο---τα___" in the lyrics editor results in

after

rather than the expected

before

Am I doing it wrong?

### LINKED ISSUE #2230: Melismas in Romanian lyrics
I spoke to someone from Romania, who gave me this feedback:

> We use the same conventions as Greeks use in their books. The Romanian language uses the dashes and underlines conventions only in western classical music, while, if you look at the Romanian psaltic chant manuscripts, they all use the prolonging vowel system that the Greeks use. Now, important note, what IS different from Greek convention is that we don't have the double vowel situation that the Greek language has, where you need to write both vowels to keep the sound. for the /u/ sound in a Greek chant you would have to continuously press tab on the [ου] syllable, while, in Romanian you just use the [u].
>
> Image
>
> Image
>
> I know that right now if I use underline in Neanes it just adds an underline like in classical music (or English chant), and it doesn't behave like Greek. It would be nice if it would continue with the last vowel of the syllable, that's the standard byzantine chant is written here. Here is an example:
>
> Image
>
> So, for Romanian, a slightly modified version of the Greek convention would be the primary way of writing in Neanes. I wouldn't eliminate the English convention, though, as […] it is easier to read for the uninitiated.
>
> It might be helpful to have yourself a look in Romanian Byzantine chant books, so I'll attach some links where you can view some old and some new written scores.
>
>    Most popular Romanian Anastasimatarion: https://stavropoleos.ro/wp-content/uploads/biblioteca-digitala/Psaltichie/psaltichie_carte_noua/psaltichie_carte_noua_teoreticon_etc/macarie_suceveanu_anastasimatar_buc_2002.pdf
>    Doxastarion: https://stavropoleos.ro/wp-content/uploads/biblioteca-digitala/Psaltichie/psaltichie_carte_noua/psaltichie_carte_noua_doxastar_etc/lacu_nectarie_doxastarulmare1_iacov_hurmuz_1999.pdf
>    Anastasimatarion manuscript in Romanian (look especially at pg. 113 onward for papadic examples): https://stavropoleos.ro/wp-content/uploads/biblioteca-digitala/Psaltichie/manuscrise_fondul_stavropoleos/ms11353_anastasimatar_catavasii_etc_StPopescu_1882_ioansangerean.pdf

Today the software follows Western conventions by default when typing Romanian lyrics, and from the above correspondence, it sounds like the Romanian user didn't necessarily want that to be eliminated outright, but considered it a secondary or uncommon use case, while the primary use case for Romanian is to follow Greek conventions, which is a missing feature today.

The current autodetection of Greek mode from the presence of the Greek alphabet wouldn't work here, as Romanian and English share the Latin alphabet. So there needs to be some encoded notion of which language is in use. A document-level setting for this is the wrong level. The setting needs to be at the paragraph level or even lower (the neume level?), since scores can mix e.g. both English and Romanian.

It sounds like the main difference between Romanian and Greek here is that with Greek dipthongs, both letters in the dipthong are extended, but with Romanian dipthongs, only the second letter in the dipthong is extended.

Note that in Romanian Byzantine music, individual syllables can contain hyphens:

Image

So I'm not sure how we would express this from a keyboard entry point of view.

### LINKED ISSUE #2458: Greek melisma issue across kentimata boundary
**This bug was found in**

- [x] Desktop
- [ ] Web App
- [ ] Both

**Describe the bug**
Having a syllable such as "ιν" or "εν" that starts before kentimata and ends after kentimata, the melisma does not continue past kentimata:

Image

I do not know if this is a valid issue, but for some scores that I type, I cannot use the melisma feature.

**Expected behavior**
I think that the expected behavior should be that the consonant at the end of the syllable should not "break" the melisma continuation at kentimata and have a behavior as this:

Image

**Screenshots or files**
The following are two samples that I have at hand from books (unfortunately, I do not have other samples at the moment):

From G. Konstantinou Hermologion Ioanou Protopsaltou:
Image

From Vatopedinon Doxastikarion:
Image

- OS: Ubuntu 26.04 LTS
- Version v0.5.39

basil:
This PR implements support for vocalic lyrics for Romanian and Slavonic. It also improves round-tripping for Greek lyrics and makes the Greek implementation more robust. This, along with Romanian and Russian localization of the UI, should open the software to a substantial number of new users throughout the world.

Romanian was by far the trickiest to implement, as its pronunciation rules are very complex. I wrote to someone from the National University of Music Bucharest, who put me in contact with a local Neanes user who has been testing these changes and kindly answered a number of my questions about how lyrics are pronounced in Romanian. That user is still testing these changes, but preliminary feedback has been positive so far. I will wait for final confirmation from that user after more testing has been done before integrating.

I was unable to find anyone to test the Slavonic changes. I wrote to a Byzantine choir in Moscow, but I am not sure if they received my message and I did not get a response. I did have AI-fuzz test against the Russian dictionary and a downloaded corpus of Slavonic liturgical texts, and I did typeset a piece from Pop Teodorov's Anastasimatarion myself, and everything seemed to work fine, but I can't verify as I don't speak the language. My friend (who is a native Russian speaker but doesn't know Byzantine music) read over an AI summary of the pronunciation rules and said it all makes sense, but couldn't confirm the details. So far no native speaker has actually tested this, but I feel the changes are probably working, or else really close to working, and we'll find out if there are any gaps.

Changes are fully backward-compatible. With this PR the docs for Greek have been updated to prefer the use of hyphens and underscores when entering lyrics, since that preserves the most information in the document for proper round-tripping and handling of tricky cases like τω ων, including allowing the user to flip back and forth between Western and vocalic modes without any data loss. But for older documents, or if users just couldn't be bothered to distinguish between - and _ when typing lyrics, the existing behavior is backward-compatible, just without the abovementioned advantages.

I didn't bother with an AI-generated summary of these changes, but if it is helpful I can generate one.

Fixes #2458
Fixes #2230
Fixes #712


basil:
# Testing notes: lyric assignment and melisma styles

## Round-trip fidelity vs. the pre-feature baseline

The lyric pipeline is a pair of functions that must be exact inverses:
`LyricService.extractLyrics` turns the notated score back into the text of the
Lyrics pane, and `LyricService.assignLyrics` distributes that text back over the
notes. Any disagreement between them shows up as text that shifts, duplicates,
or disappears when lyrics are re-assigned after notation.

The table below measures that round trip against `0.5.45` (the commit before
Romanian and Church Slavonic support), broken out by melisma style and by the
script of the lyrics. Each row is a separate campaign of randomly generated
lyrics drawn from real corpora: the hunspell `el_GR`, `en_US`, and `ro_RO`
lexicons, the AGES polytonic Greek liturgical library, and a Church Slavonic
text corpus. Words are hyphen-split into syllables and given melisma breaks at
random, then assigned and extracted repeatedly until the result stops changing.

Every generated score is given more notes than the lyrics can possibly consume,
so nothing is lost to running out of notes. "Correct" is therefore exact and
script-independent: the round trip must converge, and the letters of the final
extraction must equal the letters that were typed. That criterion does not care
whether a melisma was stored merged (`των__`) or vowel-repeated (`τω ω ων`),
only whether the text survived, so the Western and Greek rows are directly
comparable.

| Mode / script          |      n | 0.5.45 correct | current correct | improved | regressed |
| ---------------------- | -----: | -------------: | --------------: | -------: | --------: |
| Western (Greek text)   | 40,000 |        100.00% |         100.00% |    0.00% |     0.00% |
| Western (Latin text)   | 40,000 |        100.00% |         100.00% |    0.00% |     0.00% |
| Auto (Greek monotonic) | 40,000 |         12.38% |          91.99% |   79.61% |     0.00% |
| Auto (Greek polytonic) | 39,698 |         17.02% |          90.48% |   73.50% |     0.04% |
| Auto (Latin, Romanian) | 40,000 |        100.00% |         100.00% |    0.00% |     0.00% |
| Auto (Cyrillic)        | 40,000 |         99.99% |          95.16% |    0.00% |     4.84% |

Reading the rows:

- **Western is at exact parity.** Western style bypasses the vocalic machinery
  on both sides, and both are fully correct. Improvements do appear in Western
  style on scores with awkward furniture (Kentemata, ties, right-aligned
  martyria, insufficient note capacity), where a separate campaign of 149,794
  hostile scores found the current code preserving text that `0.5.45` destroyed
  in 5.6% of cases and losing text in none.
- **Greek is where the correctness gain is.** `0.5.45` round-trips a Greek
  melisma correctly only about one time in eight: its merge duplicates the
  repeated vowel and drops the syllable onset, so `των-` becomes `τωων` and then
  degrades further on each pass. Regressions are essentially absent.
- **Romanian shows parity because Auto routes Latin script to Western.** The
  Romanian vocalic analyzer engages only when Vocalic style is selected
  explicitly, which has no counterpart in `0.5.45`, so that work is purely
  additive and cannot be compared here.
- **Cyrillic is the one row where the current code scores lower.** `0.5.45`
  scores 99.99% only because it did nothing: it treated Slavonic as plain text
  and copied it through. Performing a real vocalic merge costs 4.84% imperfect
  round trips. Decomposed, text fidelity is 99.77% (only 0.23% of cases involve
  any letter difference at all) while convergence is 95.38%, so the cost is
  almost entirely mark-encoding churn rather than text corruption.

Two caveats when quoting these numbers. They measure the extract/assign round
trip, which is what happens when lyrics are re-assigned after notation, not a
single first-time entry. And the generated lyrics are melisma-dense: roughly 40%
of words carry a melisma break and 55% are split into syllables. `0.5.45`'s
remaining correct cases are essentially those with no vocalic structure at all,
so the practical reading is that it failed nearly every Greek round trip
involving a melisma.

## Known limits

Two failure classes remain and are not defects of a particular code path but of
what the Lyrics pane text can express. The pane string cannot encode
per-continuation break style on notes that carry no mark of their own, so a
melisma stored with hyphens can come back with underscores after "Assign
Accepts Lyrics" followed by a re-assignment. Measured over 100,000 scores this
affects 8.30% of round trips, against 10.29% for `0.5.45`, so it is pre-existing
and slightly improved rather than new. Separately, a score whose notes run out
mid-word can alternate between two equivalent mark encodings without ever
settling; no letters are lost, and the sung text matches `0.5.45`.

Closing either generically requires changing what the pane text can represent,
not more special cases in `LyricService`.

## Reproducing

The property-based harness used for these campaigns is not checked in; the repo
suite (`npm test`) is entirely example-based. Rebuilding it requires the
hunspell dictionaries under `/usr/share/hunspell` (note that `el_GR` is
ISO8859-7 and needs converting), a Church Slavonic corpus, and the AGES
`gr_GR_cog` library for polytonic Greek. The invariants worth reimplementing
are: analyzer structural invariants over each lexicon, a differential against
the pre-feature commit, extract/assign fixed-point convergence, and inline
typing idempotence.

Adding a small seeded round-trip property test to the repo suite would preserve
the invariant these campaigns check, which the example-based tests do not state
directly.

danielgarthur (dismissed):

danielgarthur:
I know next to nothing about Romanian, so I am of no help there. I can read Russian fluently, but I am unfamiliar with Church Slavonic, so am probably no help there either.

Issue #2471, opened 2026-07-14
User experience for initial martyriae


basil:
I've created about 400 documents with initial martyria objects, which I'd like to eventually convert to rich text boxes using English descriptions. However, the prospect of doing this manually is daunting, and I doubt AI would be of much direct assistance here (this kind of migration is too far outside the training dataset). It would be nice if a feature were available to convert an existing initial martyria object into the equivalent rich text box. The rich text box could then be edited as appropriate.

Alternatively, I kind of feel like a rich text box is overkill for my use case here, and perhaps that's why I've been hesitant to migrate. I really like the UI of the existing initial martyriae, where I can double click to choose from a nice list with descriptions, and I don't have to worry about setting various scales and playback options, that is all done for me. All I really want is to be able to write them in English (with correct grammar!) and use my own fonts for English and Greek (e.g. vertically stylized πλ) text (not necessarily the same font, though, as not all of my fonts have Greek characters or vice versa), and to be able to "inline" the mode key so that I can put a regular (rich) text box before or after it on the same line. I guess what I'm getting at is, I find the UX of "rich text box for mode key" to be too manual and cumbersome, while the UX of the existing initial martyria object is good, but not flexible enough. I wish there was some kind of middle ground.

danielgarthur:
I was having the same thought this morning. We need a middle ground. 

With AI I am more willing to experiment with this because I can fail faster if something isn't working out. So I will take a look at this soon.

basil:
A revised set of sketches of what I'd like to be able to create from a middle ground UX:

<img width="576" height="768" alt="Image" src="https://github.com/user-attachments/assets/9a19fd6c-4c70-44a4-92ce-1536de02f539" />

I'm not exactly sure which of these I would end up with, but I consider them all defensible from one point of view or another.

basil:
FWIW I thought about this some more today and will probably implement either the very first option or the HTM-style ordinal at the bottom in my own scores, in case that informs prioritization. It might not be necessary to support every single possible combination, since rich text boxes can be used for advanced cases. But I feel that the Romanian/Slavonic/Arabic style shown above is common enough in the historical examples I studied that it is probably worth having first-class support for.

PR #2522, opened 2026-07-25
feat: custom mode keys revisited

danielgarthur:


danielgarthur:
@basil I have been working on this new custom mode key feature. It is rough, not ready for a code review. The UI is incomplete and sloppy.

However, the main idea is in place. Would you mind testing it to see if it will accomplish what you think it ought to accomplish?

The basic idea is

- Go to `Format -> Initial Martyria Styles`
- Look at all the built-in styles.
- Choose one to copy.
- Customize it.
- Use it (there is a button in the dialog to use the selected style for the entire document)

For any mode key you can add any number of custom text components, as well as one of each of the following neume font components
- Echos
- Stacked "PL" plagal symbol
- Mode Sign Martyria
- Starting note cluster

Starting note clusters come in two flavors. One uses the neume font's starting note names. The other allows the user to enter their own names for starting notes. This can be used to use your own Greek font, or it can be used to transliterate the note names in English/Romanian/Arabic.

I included a number of built-in styles. I don't think I will include very many of these styles in the final PR, but they are temporarily there to aid in testing. If you have any opinions on an official list that should be included, let me know. Also, I know next to nothing about Arabic, so I couldn't test it very well. I messed around with what ChatGPT spat out for me (I think it said First Mode, Second Mode, etc. It seemed to work. But again, I don't feel comfortable evaluating the Arabic support.

Another thing to note is that I have intentionally tried to not go overboard in allowing users to customize every single thing about a signature. So for example, a user cannot override the specific font details of a single text component. It is only possible to override a component's paragraph style. So technically it **is** possible to customize every individual text components by creating a number of paragraph styles. However, this seems like an unlikely use case. I figure if users request more customization, it can be added in slowly later. 

Another thing I didn't make customizable was the spacing between various components, although I could see that as a more reasonable thing to make customizable.

An interesting oddity about the EZ Psaltica font is that the height of the capital letters (e.g. the Π in Πα) is smaller than many fonts. For this reason, I have chosen to scale the neume font in the mode keys by matching the height of capitals in the neume font to the height of the capitals in the default paragraph style chosen for the mode key. This keeps the sizing consistent.

I've left in the current mode key implementation that is in production today because I don't really want to mess with that feature. I think many users will want the simplicity of the current mode key and not want to mess with custom mode keys. In addition, the current mode key has what I assume was purposely designed kerning of sorts (in the form of bearing widths), with specific spacing between the various mode key glyphs. In order to preserve that exact look, I would have to pull that information into a built-in custom mode key and support configurable space between specific combinations of neumes. Instead of messing up the default mode key in people's existing documents, I would rather that same spacing be preserved in the simple case (since many users are not going to care about custom mode keys and will just complain about how their mode keys have changed after updating to the latest version). As to what, if anything is to be done about spacing and kerning with custom mode keys, I'm not sure, but it doesn't seem like an issue that needs to be tackled immediately.

~~PS
I'm aware of the request to allow inline mode keys. That will be a separate effort.~~

src/models/InitialMartyriaStyle.ts
```diff
@@ -92,16 +137,53 @@ export function getInitialMartyriaContext(
...
+export function createInitialMartyriaStartingNoteText(): InitialMartyriaStartingNoteText {
+  return {
+    names: {
+      [ModeSign.Ni]: 'Νη',
+      [ModeSign.Pa]: 'Πα',
+      [ModeSign.Vou]: 'Βου',
+      [ModeSign.Ga]: 'Γα',
+      [ModeSign.Thi]: 'Δι',
+      [ModeSign.Ke]: 'Κε',
+      [ModeSign.Zo]: 'Ζο',
```
ilizol: `'Ζο'` -> `'Ζω'`

src/models/InitialMartyriaStyle.ts
```diff
@@ -350,21 +350,60 @@ export const builtInInitialMartyriaStyles: InitialMartyriaStyle[] = [
...
+    'English Ordinal',
+    [
+      text('first', 'First', [1]),
+      text('second', 'Second', [2]),
+      text('third', 'Third', [3]),
+      text('fourth', 'Fourth', [4]),
+      text('fifth', 'Fifth', [5]),
+      text('sixth', 'Sixth', [6]),
+      text('seventh', 'Grave', [7]),
+      text('eigth', 'Eighth', [8]),
```
ilizol: `'eigth'` -> `'eighth'`

src/models/InitialMartyriaStyle.ts
```diff
@@ -350,21 +350,60 @@ export const builtInInitialMartyriaStyles: InitialMartyriaStyle[] = [
...
+    [
+      text('plagal-word', 'Plagal of', [5, 6, 8]),
+      text('first', 'First', [1]),
+      text('second', 'Second', [2]),
+      text('third', 'Third', [3]),
+      text('fourth', 'Fourth', [4]),
+      text('fifth', 'Fifth', [5]),
+      text('sixth', 'Sixth', [6]),
+      text('seventh', 'Grave', [7]),
+      text('eigth', 'Eighth', [8]),
```
ilizol: `'eigth'` -> `'eighth'`

basil (commented):
Wow, the visual result here looks really great. To me, this PR produces polished-looking initial martyriae and addresses the software's current limitations with an impressive degree of flexibility. My reservations are mostly about whether we can preserve these strengths with a simpler long-term design and implementation.

My guiding principles are:

- The design should be as simple as possible while accommodating roughly 85-90% of use cases. Rich text boxes can provide flexibility for the remaining 10-15% of edge cases.
- It is acceptable for output to change slightly after an upgrade, provided that it looks as good as or better than the previous output.
- A unified UI and implementation are easier for users to understand (and for developers to maintain) than duplicate systems or complex styles that must be forked and customized.

This PR allows users to define very flexible sets of styling instructions that are then interpreted to produce a layout. That is a powerful capability, and it is clearly able to generate excellent results across a wide range of examples. My concern is that these instructions are likely too complex for users to create from scratch, making the expected workflow one of copying an existing style and modifying it. Over time, that could make both the Neanes codebase and users' documents harder to maintain. (Compare to the initial draft of Header/Footer support, which started as copy/pasted headers/footers for each section and subsequently evolved into a cleaner system of document variables.)

I wonder whether we can retain this PR's strong visual results while exposing a smaller set of semantic options for the common cases. Rich text boxes could remain the escape hatch for edge case layouts. This would preserve meaningful flexibility for advanced users while keeping the normal workflow, application code, and data format approachable.

Modeling this feature as a new subsystem has been clearly useful in demonstrating what the improved design can achieve while retaining backward compatibility. For the long-term implementation, however, I would strongly prefer to incorporate those improvements by refactoring the existing system rather than maintaining parallel implementations. We already have both rich and plain text boxes, and that distinction is largely an implementation detail that users should not need to understand. Unifying these concepts would make the experience more straightforward.

The successful introduction of Knuth-Plass line breaking is a good precedent: the output changed, but the new result justified the migration by being as good as or better than the old one. I think we can take the same approach here. If an upgraded document does not look at least as good as it did before, we should treat that as a bug and improve the new implementation before shipping it. Once the new path consistently produces equal or better output, we can remove or refactor the old one instead of supporting both indefinitely. That may require more development effort initially, but the quality already demonstrated in this PR makes me optimistic that it is achievable.

For maximum simplicity, I would also prefer not to support neume font glyphs for text in the general case. Quite simply, the existing text glyphs in the neume fonts are just bad versions of Didot or Porson style Greek letters. These should be deleted from our neume fonts, and we should just ship and use high quality Didot/Porson text fonts (Didot for Psaltica, and Porson for Stathis). The typography in this PR is already compelling, and I believe we can match or improve existing documents' appearance for both Psaltica and Stathis Series with standard text fonts. Legacy documents could be migrated to GFS Didot, GFS Didot Classic, or Old Standard, while Stathis Series documents could use GFS Porson, which we would begin bundling. This would give us a single code path in which Mode, Varys, Zo, and similar text use standard text fonts.

There may be more details to work through (which I can definitely help with), but I am confident we can find strong typographic examples in professional books set with EZ Psaltica, Stathis style, and Chourmouzios fonts, then reproduce an equal or better appearance using text fonts. The vertically styled "pl." is the exception here. EZ Psaltica uses the same Didot-style font used for Hxos and Varys, whereas Stathis documents use a Porson-style font for Hxos and Varys but custom, music-like glyphs for the vertical "pl." I see two relatively simple ways to handle this:

1. Remove the "pl." glyphs from EZ Psaltica (where they are just copies of Didot letters) but retain them in Stathis Series (where they are unique music-like glyphs). Neanes could use the neume font glyphs when present and fall back to the text font when they are absent.
2. Place the music-like Stathis Series "pl." glyphs in the Unicode slots for ordinary pi and lambda. Stathis Series could then be selected and used like any other text font for the vertically styled "pl."

Either option appears capable of preserving the distinctive Stathis appearance without adding much complexity to the UI or implementation. This PR has already done the difficult work of showing what a good result should look like; this would recast that work into a simpler model.

For similar reasons, I am not yet convinced that users need to work with Initial Martyria Styles as a separate concept. This PR supports an impressive range of variations, but most of them can be expressed through a concise set of semantic options on each initial martyria:

- Language, selected from the languages supported by the application, such as English, Greek, Romanian, Slavonic, and Arabic. Users who need another language could request support or use a rich text box.
- A checkbox to show the full, redundant mode name, enabled by default for Arabic, Slavonic, and Romanian.
- A checkbox to use plagal terminology, enabled by default for Greek and English.
- A dropdown for numeral style, such as ordinal or cardinal.
- A checkbox to transliterate the note name. Each supported language should have a reasonable default transliteration, perhaps with an advanced document-level option for customization.
- Any relevant right-to-left options for Arabic.

These options would retain the most valuable capabilities demonstrated by this PR while presenting them in terms users already understand: language, terminology, numerals, and transliteration. Rich text boxes would still support exceptional layouts without requiring every document to carry copy-pasted rendering instructions.

Overall, I think this work is moving the appearance of initial martyriae in an great direction. The output is polished, a lot of attention to detail has been paid to the various use cases, and this PR provides a strong foundation for improving this part of the software. My suggestion, which I'm happy to help with in terms of design/implementation, is to build on that by narrowing the user-facing model, consolidating the implementations, and keeping one clear path for the vast majority of users. That would preserve the visual quality and flexibility of the current work while reducing cognitive load and making the feature (and users' documents) easier to maintain over time.

danielgarthur:
As always, I am all for simplification. As far as the UI goes, I would push for even simpler if we are going for simple. 

Like a dialog that shows only a language selector and then some panes that show a handful of representative signatures so the user can just see what they are getting. I don't have time right now to make a full mock up, but something like a bunch of these side by side (but that only show two or three signatures like an authentic mode, a plagal mode... maybe varys. That kind of thing.)

image

So they don't need configure ordinal vs number or plagal vs no plagal. Just look and see, "I like that one". Click. If any config is available maybe it's the option to transliterate the note names, since the panes might be redundant otherwise.

The only thing that I wonder about with this simplification is whether users will need the one-off option to use alternate text for a specific mode key template. I can't find the issue now, but someone wanted to add the Greek word for enharmonic after this signature. 

image

I'm not sure whether there may be various other examples. Maybe if someone wants to write a maqam name or something. I'm not sure if this belongs in the realm of "use the rich text escape hatch" or whether we should allow some kind of additional text to be inserted easily. 

For inline mode keys, users can just put a rich text box along side the mode key, but for full length mode keys (say a centered signature, for example), the UI needs to be well thought out for that to work well, if it is to be supported at all.

basil:
> As always, I am all for simplification. As far as the UI goes, I would push for even simpler if we are going for simple. […] Like a dialog that shows only a language selector and then some panes that show a handful of representative signatures so the user can just see what they are getting. […] So they don't need configure ordinal vs number or plagal vs no plagal. Just look and see, "I like that one". Click. If any config is available maybe it's the option to transliterate the note names, since the panes might be redundant otherwise.

That all sounds fine. The only thing to keep in mind is that once you go outside of the classical Greek tradition, the conventions are less obvious and widely agreed upon, especially in English where there is the unusual (compared to other non-Greek languages) situation of (a) a widespread desire to use plagal terminology and (b) the grammatical constraints around word order. So it might be tough for users to understand the reasoning and tradeoffs behind the various options from a picture alone. I probably wouldn't be able to easily understand the logic behind the options myself if I hadn't spent a few hours combing through old books in various languages to see what people have done in the past. In my sketch in #2471 I had to write labels to "explain" the drawing, so if we go the visual route some type of "caption" might be useful. Otherwise we might get tickets asking to be able to write "Mode Plagal 2" or similar.

> The only thing that I wonder about with this simplification is whether users will need the one-off option to use alternate text for a specific mode key template. I can't find the issue now, but someone wanted to add the Greek word for enharmonic after this signature. I'm not sure whether there may be various other examples. Maybe if someone wants to write a maqam name or something. I'm not sure if this belongs in the realm of "use the rich text escape hatch" or whether we should allow some kind of additional text to be inserted easily. For inline mode keys, users can just put a rich text box along side the mode key, but for full length mode keys (say a centered signature, for example), the UI needs to be well thought out for that to work well, if it is to be supported at all.

Usually things like "Enharmonic" aren't written explicitly as a word but rather with the ajem fthora, but yes there are use cases for various other words to be inserted like "Phrygios", the makam name, or even the generic "Fthorikos" (which Stanitsas used to note his Dynamis in Makam Ferâhnâk). [This paper](https://pergamos.lib.uoa.gr/uoa/dl/object/2867197) has some examples of the more wild and whacky ones. In any case I don't see a real distinction from a UX perspective between the above and non mode key text, like composer name, "Both now. Theotokion" or "Plagal of Fourth Mode. By conceiving" and other such rubrics. It is extremely common to write additional rubrical text besides an initial martyria, so this should be a first-class use case. While I suppose this could be done with the initial martyria object, it might end up just becoming a rich text box at that point, so maybe we want to encourage a combination of rich text box + inline martyria side by side, making sure that use case works well and can be centered properly, etc. Or alternatively, allowing the current fully-featured initial martyria object to be "embedded" in a rich text box, like an image, but I'm not sure how difficult that would be to implement or whether it's worth the complexity.

BTW the juxtaposition of Source Serif with Didot in

image

didn't look good to me, in part because these are two very different serif styles but also because Irene Vlachou did such a great job designing the Greek version of Source Serif, so it's a pity to lose that in this example.

The currently shipping software produces

image

which I think looks horrible, a combination of Porson, a classic looking Didot, and some really poorly digitized Didot for the note name. This would all look much better if we got rid of this from the neume fonts, shipped the GFS Porson text font, and then used it to render Stathis-style scores. For the Chourmouzios font, the Greek glyphs in EB Garamond would probably go rather well.

The closest high quality text font I can find for the EZ Psaltica Greek glyphs are either GFS Didot Classic, or Old Standard. We already ship the "regular" GFS Didot, which has a slightly higher x-height to go with Palatino, but I think we could retain compatibility pretty well by using GFS Didot Classic, or if bundling yet another font is not desirable then Old Standard (which we already bundle) seems perfectly fine as well.

danielgarthur:
@basil I am becoming lost in the various possibilities again... Just to support your list here https://github.com/neanes/neanes/issues/2471#issuecomment-4972082956 would require a lot of knobs. I think there are yet more reasonable possibilities, as well, for English. Whether you use "plagal of first" or "plagal of **the** first". Whether you use "tone" instead of "mode", which is a more direct translation of the Greek and Slavic. Writing out "Five" instead of "5". Probably other cases, too.

I'm remembering again why I created the extremely customizable dialog. However, I still think that supporting a curated list is probably the way to go. I just need help creating a focused list. That could very well mean that there are 20+ options for English (maybe listed all out as example panels for the user to choose from or maybe gated behind knobs or filters). Or maybe we only support a limited subset unless someone requests another. I've also considered the possibility that there might need to be different knobs for different languages.

So what I'm saying is I'm feeling overwhelmed by the choices in English, as well as my lack of knowledge about conventions in other languages. I know you've done some research here. If you could help reel me back in, I could hopefully finish this feature out.

In the latest commit, the customizable styles are gone and there is now a rough style choosing dialog.

basil:
@danielgarthur My guiding principle is that when there is clear community consensus (~90% of usages), we should support the consensus and leave the other 10% to rich text boxes, but where usage is evenly split, we need to support all reasonable/defensible choices.

We can cover 90% of cases with just 5 semantic pieces of data at the model layer (not necessarily UI elements):

1. Language drives the translation of Hxos/Varys and sensible defaults and should probably be a UI select or tab navigation, as this is the most basic starting point. For English, "Mode" is the clear consensus, as it is more accurate per the Harvard Dictionary of Music and used by 90% of serious Byzantine composers in English (even though liturgical books often use "Tone"). "Tone" is 10% territory in my opinion. For other languages Glas / Glasul etc. is clear-cut.
2. On numeral style (e.g. ordinal vs. cardinal, 1st vs. First, One vs. 1, IV vs. 4) usage is evenly split. Numbers are a universal part of the human experience, and everyone has their own personal preference. I think this needs to be customizable, e.g. like this [Word number style menu](https://erinwrightwriting.com/wp-content/uploads/2022/03/Word-365-Define-New-Number-Format-Dialog-Box-Number-Style-Menu-copy.jpg).
3. On transliterating note names usage is evenly split. Older Romanian/Slavonic books tend to transliterate, while more modern ones tend not to (perhaps due to the popularity of EZ Psaltica). In any case, we need to support both, and this depends on 1. Whether this is a checkbox or inferred from language is an open question, but fully customizable transliteration is 10% territory in my opinion.
4. On using Plagal terminology, there is a clear consensus by language: it's always used in Greek, it's used 90% of the time in English among serious composers (even though liturgical books often use the Slavonic system), and it's usually *not* used in Romanian/Slavonic/Arabic. Sp this also depends on 1. So we can likely infer this from language, or perhaps add a checkbox. "Plagal of *the* First" is 10% territory in my opinion.
5. "Redundant?" as in my drawing is never the case in Greek but almost always the case in Romanian/Slavonic/Arabic. English is largely uncharted territory, so we need to support both I think. Implementation of non-redundant for English depends on 1 and 2, from which the first few drawings of mine can be built. For Greek/Romanian/Slavonic/Arabic this could possibly be inferred from language.

With 1-4 you can build the bottom "redundant" half of https://github.com/neanes/neanes/issues/2471#issuecomment-4972082956 for any language: full localized mode name [with chosen numeral + Plagal style] + Greek martyria without the "mode" word [with chosen transliteration]. That covers most historical examples in languages other than English. 5 then enables my "native" ordinal/cardinal idea for English. This is fundamentally 5 concepts, so at most 5 selects/checkboxes, which I feel is reasonable compared to the dialog we started with. And 3-5 can be auto-preset by language, e.g. Greek / English -> Plagal checked.

danielgarthur:
@basil I've implemented a hybrid approach: keep the preview panels, but also include filters to narrow things down. It's still in a rough state, just to prove out the basic UX. Let me know what you think.

Also, at this point, if you have a list of styles we should bundle that are not currently included, that would be appreciated. Feedback on currently implemented styles would help as well.

I know nothing about Arabic or its conventions. I'm sure what I have is wrong and will need your input or maybe we need to find someone who can review our work. I'm fine with leaving out Arabic for now if we aren't sure the correct conventions.

I'm also unsure about Romanian. I have seen, for example, stuff like Glas 7 <mode sign for varys>, as well as Glas 7 varis (or whatever the Romanian transliteration is) <mode sign>, not really sure what's standard there.

In Greek I have seen things like ηχος α',  not sure if we want to include those, too.

basil:
> @basil I've implemented a hybrid approach: keep the preview panels, but also include filters to narrow things down. It's still in a rough state, just to prove out the basic UX. Let me know what you think.

This is really awesome, it is far more intuitive for me. I played around with it and I can get some great results. Setting the Greek text font to Old Standard yields something pretty close to what we're shipping today, as expected, but the note name is a lot smaller than the word "Hxoc" - perhaps that is just a temporary condition since this code is rough and under development.

> Also, at this point, if you have a list of styles we should bundle that are not currently included, that would be appreciated.

I have a pretty large PDF library of Byzantine music books in all kinds of languages, including older ones as well as modern digitally typeset ones. I will go through some of them over the next few days and check the conventions used and if they can be represented in this new system and let you know if I find any gaps.

> Feedback on currently implemented styles would help as well.

My feedback is very positive overall, I find the UI about as simple as it could be for such a complicated issue. The only thing that I find odd or inconsistent is the English descriptions "Sign First", "Plagal First", or "Mode Before Sign". While these are all technically true, they describe a particular presentation or implementation of a fundamental concept, without stating the underlying logic or fundamental concept itself, which is that the ordering depends on how this would be pronounced when reading a particular numeral style out loud. So, for example "Sign First" would be read out loud "First Plagal Mode," and "Mode Before Sign" would be read as "Plagal Mode 1". The thing currently labeled "English - Plagal First" would be read out loud as "Plagal of First Mode." These orderings are effectively an implementation of numeral style, at its heart a very personal and subjective preference. But this isn't clear from either the captions/labels, or the fact that they disappear when I select a numeral style from the dropdown. On the contrary, selecting a numeral style should show _both_ variations for that numeral style, the historical "redundant" one, and my proposal of a native one. In other words, today when you select "1" as the numeral style and "Use plagal terminology" is checked, you only see what is currently labelled "English - Plagal Mode 1", but I would also expect to see what is currently labelled "English - Mode Before Sign" -- because, if that were to be pronounced out loud, it would be pronounced (I suppose) "Plagal Mode 1", so it is effectively an encoding of a particular numeral style preference, just without redundancy. I think it is important for this concept to be cleanly modeled in the UI, because otherwise users will have no idea what any of this represents, and then the first ticket we will get after this ships will be a request to add "Mode Plagal 1."

> I know nothing about Arabic or its conventions. I'm sure what I have is wrong and will need your input or maybe we need to find someone who can review our work. I'm fine with leaving out Arabic for now if we aren't sure the correct conventions.

I am originally from Lebanon, but I grew up in the United States and don't speak Arabic. If needed I can ask questions to my family members who are native speakers, but they don't know Byzantine music. I can look through the PDFs I have in Arabic by Mitri el-Murr, Fr Nicholas Malek, Joseph Yazbeck, Hamatoura, and Melkite sources as part of the abovementioned corpus check, but I can't promise any particularly deep insights here. For those, a native speaker who also knows Byzantine music would be needed.

> I'm also unsure about Romanian. I have seen, for example, stuff like Glas 7 , as well as Glas 7 varis (or whatever the Romanian transliteration is) , not really sure what's standard there.

Similar to the above, I will check the old books I have by Suceveanu, Macarius, Anton Pann, as well as modern reprints of them, and modern PDF scores, but I can't promise any particularly deep insights.

> In Greek I have seen things like ηχος α', not sure if we want to include those, too.

Do you have an example? I guess it couldn't hurt to allow redundancy even for Greek, but as far as I understand this isn't common at all, so I'd be curious to know what you're referring to.

danielgarthur:
Thanks. Much of the text like the captions is AI slop that I wasn't wanting to fix until a path was chosen. 

Kypsele uses that notation for sure, the 19th century publication. I have seen it elsewhere in Greek publications, I am sure. Papa Ephraim included it in EZ Psaltica and it appears occasionally in his Vespers and Orthros books.

I'll ask around on my end regarding Arabic and Romanian.

basil:
> Kypsele uses that notation for sure, the 19th century publication.

Oh right, yes indeed I see what you mean. Yeah that should probably be supported as yet another numeral style.

I have a nagging feeling that one of the reasons I keep seeing the same conventions in modern reprints that use EZ Psaltica or that were typeset with Melodos is that those conventions were based on technological limitations or ease of use rather than deliberate. In older books I see far more deliberate conventions (and far more variation, as you would expect for personal preferences like numeral style).

HAyoub33:
Hi, first thank you for this tremendous and nice job! appreciated to have neanes for most of the feasibility it has.

I would help for what is related in arabic. Let me precisely know what do you need. 

Newly working on Neanes, and i find the customized key signatures is a bit complicated to build (or something i m missing), even by trying the example file available on github. 
Modifications in the properties look weird and unresponsive. I remember in the past when checked on Neanes, was the ability to add initial key signatures and give a name for it (like a makam) in the same box where the preset Signatures are available. 
Is it possible to go back to that kind of feature ?

Meanwhile, any help is appreciated for how to build a custom key signature withing the inline rich text box, specially using different keys  in the same file, (e.g. psalm with multiple keys).

As new also on github and on your system, and if this is not the correct place to ask these questions please guide me to the correct place. 

Thank you

danielgarthur:
@HAyoub33 Thank you. I've created a place for us to discuss this: https://github.com/neanes/neanes/discussions/2567

basil:
@danielgarthur BTW I did some more experiments with NeanesTeX and I am leaning toward a convention like this for my print edition(s):

image

image

image

image

I'm still undecided if I want to include the redundant ancient symbols for the mode name. On the one hand, most 19th century Romanian and Slavonic publications do include them, with all the redundancy that entails, but on the other hand there _is_ past precedent for dispensing with the ancient mode name symbols in _Mousike Kypsele,_ as you showed me earlier. So far I like the examples above because they allow me to blend the conventions of liturgical books with the conventions of music books, which is particularly useful for publishing modern complete editions that have both the service text and music.

I kept the Greek in the above example to be consistent with the use of Greek in (non-initial) martyriae. Since Heraklit doesn't include italics, I used Source Serif 4 Variable Italic for the Greek and tuned the weight to 500 to match the weight of Valkyrie. I think the result looks quite nice from a font harmony perspective.

Whatever I decide on in the long term for the print edition(s), I will likely backport it to the free PDFs on my web site.

basil:
While Manasia Pop Teodorov's books (certainly among the most respected in the Slavonic tradition) do include the redundant ancient symbols for mode names, some others don't. These two books in Slavonic are a bit inconsistent, sometimes they include the redundant ancient symbols for mode names, and sometimes they don't:

Paschalia by Joan Harmosin-Ohridski, 1869:

image

Strasna Sedmica by Ilia Atanasov, 1941:

image

The only thing I can conclude with confidence is that in 19th century publications in Romanian and Slavonic there is historical precedent for both including and not including the redundant ancient symbols, with the majority including them.

basil:
@danielgarthur I took a pass to recast this change as semantic metadata: numeral kind (cardinal or ordinal), numeral style (digits, roman numerals, alphabetic numerals, or words), numeral placement (postnominal or prenominal), plagal terminology, and mode identification method (text only, mode signs only, or both text and mode signs). The hand-authored component lists are now language lexicons with a centralized algorithm to generate the components from the language lexicons. This effectively replaced the hard-coded implementations for each built-in style with a centralized implementation that covers the cases for all languages, based on a language lexicon. This gets to the essence of the "problem" we are trying to solve here.

I also took a first pass through Romanian and some Slavonic sources to look for attested usages that were not covered as built-in styles. I added 8 built-in styles across English, Spanish, Church Slavonic, Russian, and Romanian for usages I found. I also started annotating the built-in styles with sources that attest them, or explicitly annotating them as unattested if they are unique to Neanes. I still want to take a second pass over these to annotate more built-in sources, as well as to check more books I have in PDF form in Slavonic and Arabic to see if there are any other missing but attested variations. I will probably take another attestation pass, but unfortunately it can take some time and effort for me to dig into these sources, as I am not very familiar with these languages. For example I want to look at some Melkite RTL PDFs I have.

Along the way I improved punctuation, wording, ordering, and plagal/grave-mode presentation, and I made transliterated note names the default for Church Slavonic and Russian. I also expanded the style dialog with “All languages” and filters for numeral kind, numeral style, terminology, and identification method and added automatic handling for incompatible filter combinations and unavailable options.

Apart from checking more sources, annotating more attested built-in styles, and adding new built-in styles, etc... I think the main action item from my perspective is to work on the on-disk format and UI. Currently the on-disk format and UI are centered around a hard-coded list of built-in styles with IDs, and it isn't possible to generate arbitrary combinations. However, with my changes to model things semantically, I think it should be possible (with a little more polishing) to open up the UI to arbitrary combinations, but expressed in a maintainable way: as semantic properties, not hard-coded instructions. I think that should allow us to change the on-disk format to store the semantic properties directly on disk rather than the hard-coded IDs, which should make the software easier to maintain as well. I still like your idea of the UI giving you a bunch of options, where it is easy to visually find the one you like and select it. So I think we should retain that, but it should be something like a set of attested semantic option presets rather than the only allowable choices. Options that are not well-attested but still legal could still be produced, just maybe without an explicit preset. The current code is about halfway there, having been refactored around cleaner semantics, but the UI and on-disk format is not, so I'm hoping you might be able to pick up where I left off.

basil:
image

I added a pronunciation to each mode key, which should help improve the UX by elucidating why a particular legal combination is constructed the way it is.

basil:
> I think the main action item from my perspective is to work on the on-disk format and UI.

@danielgarthur Tried this out this morning, let me know what you think.

basil:
@danielgarthur I cleaned up the model enough to the point where I felt comfortable ripping out the old implementation and migrating to the new on document load. This involved bundling GFS Porson to maintain the distinctive look of Stathis Series scores. I also refined the model and added tests to verify the grammar/lexical rules against the Unicode Common Locale Data Repository, with specific (attested) exceptions for liturgical languages like Slavonic. That should give us a lot more confidence in our ability to support unfamiliar languages.

I still need to take another pass through the default built-in styles and either attest the remaining ones or remove them from the defaults list. Slavonic and Romanian in particular need a second attestation/cleanup pass, but the primary supporting infrastructure should be firm now, so it's just a matter of taking another pass through the sources and keeping the well-attested variations and removing the others from the default built-in styles.

I'm starting to realize that "Tone/Mode" is perhaps a bigger problem than I initially thought. I seem to be finding the same problem in other languages, like Romanian, where there seem to be two conventions, an older one by Anton Pann that uses words like "Glas" and "Laturas", while a newer Hellenophile convention that uses words like "Ehul" and "Plagal". I think I might need to add an axis/dimension to the model for alternative words for Mode/Plagal. After having modeled all the other complex stuff, this shouldn't be impossible. But I think it is a bigger problem than I first considered. I don't think the Romanian matrix can really be complete without it.

Other than that, the whole diff probably needs some polishing in the font/rendering aspects, this isn't my forte so I am not planning to work on that soon. The UI could probably be polished too; although it is already a lot better than it was before, it is still not what I would consider production-ready.

The UI now allows you to create any legal combination, where "legal" is defined as "grammatically correct in the given language." This involved building out a small lexical/morphological/grammatical layer. With that layer built out, everything else models cleanly and the resulting system is logically consistent. And yes, "Mode Plagal 2" is not grammatically correct:

> For the intended meaning (the member numbered 2 within the class of plagal modes) the ordinary English form is "Plagal Mode 2."
> 
> First, *plagal* modifies *mode*. Merriam-Webster classifies [*plagal*](https://www.merriam-webster.com/dictionary/plagal) as an adjective and records [*plagal mode*](https://www.merriam-webster.com/dictionary/plagal%20mode) as a noun. This follows the normal English order in which an attributive adjective precedes its head noun. Huddleston and Pullum distinguish this ordinary pre-head use from the restricted postpositive use and directly contrast *a young leader* with the ungrammatical *a leader young* (*The Cambridge Grammar of the English Language*, 2002, [pp. 528-529, 552](https://doi.org/10.1017/9781316423530.007)). Matthews gives the parallel contrast *a tall chief* versus the normally excluded *a chief tall* (*The Positions of Adjectives in English*, 2014, [p. 10](https://doi.org/10.1093/acprof:oso/9780199681594.001.0001)). The descriptive category is therefore *plagal mode*, not *mode plagal*.
> 
> Second, *2* is an identifier, not a quantity: it identifies one mode rather than counting two modes. Reynolds documents this use of cardinal numbers in expressions such as *Room 101* and *Highway 407* (2026, [section 5.2](https://doi.org/10.1017/S1360674325100518)). Applying the same pattern, *2* follows the complete category it identifies: [Plagal Mode] [2]
> 
> Combining ordinary adjective-noun order with post-category identification gives "Plagal Mode 2." "Mode Plagal 2" instead places the bare adjective after its noun, a position the cited grammar does not license for this reading.

danielgarthur:
This is a very complex task, and I'm struggling to communicate all that needs to be communicated, so I'll just start with bits and pieces as I have time.

Regarding the removal of the glyph-based EZ Psaltica initial martyria, you have my support in doing so, but I want to reduce the backlash as much as possible. It's such a classic look that people are accustomed to and I don't want an uproar when it's gone. I'm speaking of the EZ Psaltica version. The Stathis Series version has problems with strange font mixing, as previously noted, so I have no problem with seeing that one go. I do worry that longtime users of Neanes or EZ Psaltica may want some visual continuity between their new and existing (already published) scores, and removing the EZ Psaltica version would make that difficult to preserve.

The things that I noticed while I was working on making the custom Greek signature "close enough" were that

1. The neume font looks "thicker" or "bolder", especially the plagal symbols (P and L)
2. The alignment of the P and L is slightly different I think. I'm not sure what the right algorithm is there for centering. It may be fine as is, but it's worth experimenting.
3. Bou using the ligature for "ou" in EZ Psaltica. GFS Didot does include that ligature, so we may need to find a different font to truly preserve the original look. I had in mind that we could include an option to "use ligature if available" in the settings. That way people could choose whether they want that look or not.
4. There is a sort of kerning to the EZ Psaltica signatures (intentional or not, I do not know). This is not preserved in any of the custom signatures. Whether it needs to be, or even could easily be done, I don't know, since it probably depends partly on the font used for the non-neume text. It's not a blocker in my opinion.
5. There is a wider question of how much space should be between the neume symbols in the signature. Publications differ, but more space, rather then less seems to be the general trend. But I have seen publications where it just uses what appears to be normal word-spacing between the word "mode" and the neume symbols. I'm not sure whether this should become an option for users to choose.

The other thing of note is that I've been stuck on the inline support. I may rip it out for the first PR. I think that if it is supported, it would ideally be part of the rich text box, and the martyria element would not have its own inline variant. But I don't want to double implement the rendering logic as spans in a widget unless there is no better option. I've been meaning to experiment with whether a Vue component can be mounted inside CKEditor as a widget. I think it will be easier to work through vertical alignment issues if it is part of the rich text editor.

basil:
> This is a very complex task

Yeah it is surprisingly complex for "just" a heading at the top of the piece. The various historical inconsistencies certainly reflect that.

> Regarding the removal of the glyph-based EZ Psaltica initial martyria, you have my support in doing so, but I want to reduce the backlash as much as possible. It's such a classic look that people are accustomed to and I don't want an uproar when it's gone. I'm speaking of the EZ Psaltica version. The Stathis Series version has problems with strange font mixing, as previously noted, so I have no problem with seeing that one go. I do worry that longtime users of Neanes or EZ Psaltica may want some visual continuity between their new and existing (already published) scores, and removing the EZ Psaltica version would make that difficult to preserve.

Agreed, I think it should totally be possible to make the new version equal to if not better than the old. The biggest thing we can do to preserve compatibility here is to ship "GFS Didot Classic" and use it in place of "GFS Didot." It is a closer match to Psaltica and should mitigate backlash.

> 1. The neume font looks "thicker" or "bolder", especially the plagal symbols (P and L)

That makes sense typographically for the same reason as "fake" small caps are deficient compared to true-drawn small caps: https://en.wikipedia.org/wiki/Small_caps#/media/File:True_vs_Scaled_Small_Caps.svg

And as I mentioned in a previous comment, Stathis Series scores use a different glyph for the P and L, it's not Porson. So I think we should have an option to use the symbols from the music font for these (probably the default, for compatibility) and allow you to check a box to use the glyphs from your own Greek font. Then it would be up to you to find a matching weight (e.g. Medium or Semibold).

> 2. The alignment of the P and L is slightly different I think. I'm not sure what the right algorithm is there for centering. It may be fine as is, but it's worth experimenting.

Seems plausible.

> 3. Bou using the ligature for "ou" in EZ Psaltica. GFS Didot does include that ligature, so we may need to find a different font to truly preserve the original look. I had in mind that we could include an option to "use ligature if available" in the settings. That way people could choose whether they want that look or not.

Wow, a Greek glyph that isn't in GFS Didot or in Old Standard. But not to worry, a quality Didot-style version is tucked away in Garamond Libre (which not many people know has a Didot-style Greek set of glyphs). Honestly I think we can likely dispense with it though. Even Melodos doesn't support this:

image

> 4. There is a sort of kerning to the EZ Psaltica signatures (intentional or not, I do not know). This is not preserved in any of the custom signatures. Whether it needs to be, or even could easily be done, I don't know, since it probably depends partly on the font used for the non-neume text. It's not a blocker in my opinion.

As long as it looks good compared to 19th century editions I think we're OK.

> 5. There is a wider question of how much space should be between the neume symbols in the signature. Publications differ, but more space, rather then less seems to be the general trend. But I have seen publications where it just uses what appears to be normal word-spacing between the word "mode" and the neume symbols. I'm not sure whether this should become an option for users to choose.

In 19th century Victorian typesetting there was an obsession with adding lots of space, like adding two spaces between sentences. Modern sensibilities have strongly resisted this. I think a standard normal word space should be placed between all text words. Between text and music possibly an en space could be used, or else a regular space (erring on the side of a regular space), whatever looks better. I don't think we need to overthink it more than that.

> The other thing of note is that I've been stuck on the inline support. I may rip it out for the first PR. I think that if it is supported, it would ideally be part of the rich text box, and the martyria element would not have its own inline variant. But I don't want to double implement the rendering logic as spans in a widget unless there is no better option. I've been meaning to experiment with whether a Vue component can be mounted inside CKEditor as a widget. I think it will be easier to work through vertical alignment issues if it is part of the rich text editor.

Yeah I think it would be ideal if we could use rich text box for this and "embed" the mode key inside it and have it line up with whatever additional text the user typed.

basil:
> There is a wider question of how much space should be between the neume symbols in the signature. Publications differ, but more space, rather then less seems to be the general trend. But I have seen publications where it just uses what appears to be normal word-spacing between the word "mode" and the neume symbols. I'm not sure whether this should become an option for users to choose.

I think a user option is overkill. We should just design something sensible, it shouldn't be that hard. Victorian-era massive spacing doesn't appeal to modern sensibilities. Between any normal text, a regular space should be used, including after a period like "Plagal of Fourth Mode. Ni." The only place where this gets tricky is between musical characters, or between a text character and a musical character (or vice versa), since I'm not sure what size a "normal" space should be there. Possibly it should be a bit bigger because the music font is like size 20 instead of the text font which is size 14.5. I tried taking a look at the existing code but I got dizzy trying to tease apart the combinations. Anyway, apart from slightly larger spacing around music than around text, I don't see a need for the large half-em or en-sized spacing of the 19th century publications.

basil:
image

Did a UI touchup/facelift with Fable 5.1, I think it looks alright. Feedback welcome.

basil:
While I think the dialog box is starting to look OK, the workflow is still a bit awkward / weird. I'm not sure offhand how to fix it. The names of default combinations leave a lot to be desired and the idea of a user "naming" one of these styles seems inherently awkward to me. Unlike, say, a paragraph style, these styles don't really have any semantic meaning. Moreover, any attempt to name them after some existing setting would quickly become out of date if you change that setting. Also unlike paragraph styles, there aren't an infinite number of combinations... with all supported languages and axes + grammar rules, there are about 250 legal choices. So creating a "new" one isn't really an act of novel creation, but more of "bookmarking" one of the 250 choices for you to be able to access it easily in the future without exploring the whole space. There's probably a better way to do this...

basil:
> There is a sort of kerning to the EZ Psaltica signatures (intentional or not, I do not know). This is not preserved in any of the custom signatures. Whether it needs to be, or even could easily be done, I don't know, since it probably depends partly on the font used for the non-neume text.

I took a closer look at the kerning. GFS Didot Classic is missing about 200 kerning pairs from GFS Didot, this seems to be an oversight. There's been an [Ubuntu bug report](https://bugs.launchpad.net/ubuntu/+source/fonts-gfs-didot-classic/+bug/1739316) open for a decade. It would be trivial to use AI to copy over the missing kerning pairs though.

(I wanted to check if a "professional" font had a better example here, but shockingly, the gold standard Greek Didot in historical printing, Monotype Greek Series 90, has never been digitized! How shocking is that!)

In any case, I think GFS Didot and GFS Didot Classic are the best digitizations we've got for Greek Didot, and the glyphs really do look good to me and are faithful to Fermin Didot's letters (even more so than whatever is in EZ Psaltica), so I think this is indeed an upgrade. But some kerning pairs do need to be fixed or added, like between Hx in "Hxoc" and between the Ga in "Ga", etc.

I've emailed Antonis Tsolomitis to see what he thinks about all of this. Perhaps he's willing to add the missing pairs and release a new version of GFS Didot Classic, which is otherwise working really well I think. If he doesn't respond, we could always fork it and add the pairs ourselves.

danielgarthur:
I was actually referring to the space between the word echos and the mode martyria. I can send an example later. But if we go the route of using a word space then it doesn't matter much, perhaps.

basil:
I see. Well I do really think GFS Didot [Classic] can be improved a bit on the kerning front, and will try to work with the maintainers to make some targeted fixes for our use case, but this isn't a blocker IMO.

basil:
Oh, and the _Heirmologion_ of John the Protopsaltes doesn't use the "ou" ligature either, so I don't feel too bad about dropping it.

basil:
@danielgarthur OK, I did some digging and found the exact Greek font used in EZ Psaltica for the glyphs. It's "SKAmaliaOldPolUni_W", available in https://web.archive.org/web/20120220050906if_/http://enoriaka.gr/fonts/enoriaka.gr_fonts_a.zip

> Copyright (c) 2003 Spiros Krontiris.  Politonal [sic] fonts for Word with operate [sic] system Windows 2000 and Windows XP. e-mail: spikro@otenet.gr

Honestly, the font is a complete mess, it seems to be a legally dubious ripoff with various glyphs from Palatino thrown into the mix. I wouldn't feel comfortable bundling it in Neanes.

But if anyone wants the _exact_ same output they had before, this can be an escape hatch: we can tell them to assess its legal status and potentially download it and set it as their Greek font for initial martyriae. If they do that, they will get the same output as before.

This makes me a lot more comfortable changing the default font. The only question is whether to change it to "GFS Didot" or "GFS Didot Classic." Both are in my mind higher quality fonts and under a clearly open license.

GFS Didot in my opinion looks better, it's a modern refresh of the Didot style done by Takis Katsoulidis. Even Robert Bringhurst mentioned it in his book. It harmonizes better with newer fonts with a bigger x-height, including Source Serif. But... GFS Didot Classic is closer to the old look, so it might be preferred for compatibility purposes. To a naive person, it might be indistinguishable from the old "SKAmalia" in EZ Psaltica, but I can certainly tell the difference.

Do you have an opinion here, how does this strategy sound and if so would you prefer GFS Didot or GFS Didot Classic? They're both in the branch right now in case you want to do an A/B test for yourself.

danielgarthur:
I think GFS Didot looks better.

PR #2524, opened 2026-07-26
fix: use exact printing

### LINKED ISSUE #2369: Rich text box missing pictures when printing
when i add any picture on rich box and want to export it as a pdf it doesn't show on the pdf file

### LINKED ISSUE #2523: Drop Cap fill rendered incorrectly in PDF
**This bug was found in**

- [x] Desktop
- [ ] Web App
- [ ] Both

**Describe the bug**
I have created a Drop Cap style with white fill and a red outline. While it is rendered correctly in Neanes, when I export the document to PDF, the white fill is rendered as gray.
Is this a bug?
Thank you!

The settings
Image

The Drop Cap in "Neanes"
Image

The Drop Cap in PDF

Image

danielgarthur:
Fixes #2523.

Maybe fixes #2369, although I was never able to reproduce that issue.

Uses `exact` printing. See [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/print-color-adjust).

### DIFF
--- a/electron/main/index.ts
+++ b/electron/main/index.ts
@@ -899,6 +899,7 @@ async function exportWorkspaceAsPdf(
             args.pageHeightInches,
           ),
           landscape: args.landscape,
+          printBackground: true,
         });
         const newPath = replaceExtension(args.filePath!, 'pdf');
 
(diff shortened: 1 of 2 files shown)


ichnilatis-gr:
Thank you a lot!

PR #2580, opened 2026-08-10
Remove "space after" feature

### LINKED ISSUE #2569: Negative Space After No Longer Works
the last version  of update 5.51 the space after doesn't work on it

thanks

basil:
Fixes #2569 

Remove the per-element "space after" property, which let the user add extra horizontal space following a note, martyria, or tempo element. This feature was originally a workaround for problems that don't (or shouldn't) exist. If this feature is still being used, that is evidence of a deeper bug that should be solved separately rather than worked around through this feature. This PR deletes the model fields, the properties pane controls, the `toolbar.common.spaceAfter` string in all five locales, the save/load plumbing, and all layout service handling. Existing scores that contain the field still load: every `Load*_v1` copies field by field, so an unrecognized `spaceAfter` is simply ignored. The writer emits only current fields and the score version is not bumped.

This also partially reverts commit 1ef729298e82733dad62c17a35944adcdb7db83a. The geometry helpers that commit introduced (`getNoteBoxAdvance`, `getMartyriaBoxAdvance`, `getRightInkOverhangAfterSpace`, and the `leftAdvance` parameter naming) existed only to separate a box's layout advance from its width, a distinction that `spaceAfter` created. With `spaceAfter` gone, advance is width again, so those helpers collapse. `getMeasureBarOwnerWidth` keeps using `getMartyriaBoxWidth` rather than returning to `owner.width`, so the stale-width bug that commit 1ef7292 fixed does not come back; the rest of that commit is untouched.

**Bonus:** The martyria vertical offset control was borrowing spaceAfterMax for its bounds, so it needed a bound of its own. It now uses the page height rather than the page width, matching the mode key's height adjustment. This widens the range on Letter from +/-612pt to +/-792pt.


ilizol:
@danielgarthur 
@basil 

Since now the software doesn't obey the negative "space after" property value, there is no need to keep it.
For the old Neanes scores (with negative "space after" property value), we need to update them with zero "space after" property value to be aligned with the Neanes GUI.

danielgarthur:
Conceptually, I see this as akin to MuseScore's `Leading Space` or Dorico's `Note spacing`. At the same time, I'm not sure its usefulness justifies the maintenance cost. But I am worried about the flurry of issues that may come with getting rid of it. I'm sure people are using this in fantastically inventive ways that we cannot foresee.

basil:
Well we should make a decision one way or another: either remove it and the buckets of questionable code that support it, and then deal with the user complaints to file proper issues for the real root causes (more painful in the short term, but worth it in the long term), or decide that we want to keep "space after" but only for positive widths (which means fixing the UI to not allow negative values), or decide that we want to keep "space after" fully supported for all values (which involves fixing the code I changed a few weeks ago that accidentally stopped supporting negative values).

danielgarthur:
I can't think of a good reason to support negative `Space After` values, other than working around bugs that should probably be addressed directly. I understand that some users want that level of fine-grained control over spacing, but I'm not particularly interested in maintaining it.

Positive values make more sense to me, although even there, text boxes can serve as spacers when manual adjustment is necessary.

Can you think of any compelling reason (not just for yourself but putting yourself in the minds of users) for keeping positive `Space After`? I'm trying to anticipate the feedback we will get.

basil:
Not offhand, I don't use the feature at all anymore and I think the software has improved fantastically in recent months with regard to spacing and line breaking.

PR #2627, opened 2026-09-03
feat: martyria neume select

### LINKED ISSUE #2562: Fthora placement
**This bug was found in**

- [x] Desktop
- [ ] Web App
- [ ] Both

**Describe the bug**
Write a line of music, place a martyria, align it right,  

Image

then add a neume from the list available, then add a Fthora. This last is placed above the martyria and not the new neume which is supposed to change the scale. 

Image

- OS: Windows 10
- Version v0.5.49

danielgarthur:
Fixes #2562.

## Summary

Adds full fthora support for quantitative neumes attached to right-aligned martyriae, including toolbar targeting, notation analysis, layout, persistence, and export.

## Motivation

A right-aligned martyria can contain an attached quantitative neume, but fthoras previously applied only to the martyria itself. Users need to select either component and place the fthora on the intended musical position without introducing collisions, incorrect line breaks, or right-margin drift.

## Major Changes

- Adds a martyria toolbar component selector for choosing between the martyria and its attached quantitative neume.
- Supports the standard top → bottom → remove fthora cycle on the attached quantitative neume.
- Updates notation and audio analysis so the quantitative interval is applied before its fthora changes the scale.
- Persists the attached fthora and chromatic-note metadata in the existing backward-compatible score format.
- Measures the complete martyria composite, including attached marks and surrounding tempo signs, using glyph ink bounds.
- Matches the spacing used by equivalent standalone martyria, quantitative-neume, and tempo elements.
- Positions right-aligned martyriae entirely during phase-one Knuth–Plass layout, including paragraph-start cases.
- Keeps the martyria toolbar preview synchronized with property changes and uses OpenType composition for its note and root sign.
- Updates HTML, LaTeX, and MusicXML export paths to preserve the new fthora behavior.
- Keeps the LaTeX schema at version 3 because the new fields are optional and additive.

## Implementation Notes

- The attached fthora is represented separately from the martyria’s primary fthora because it occurs after the quantitative interval and can therefore establish a different scale state.
- Shared ink-overhang calculations determine the required internal spacing between the martyria, quantitative neume, fthora, and right tempo sign.
- Paragraph-start right alignment uses a zero-width anchor and prohibited breakpoint so the leading stretch glue participates in phase-one line measurement. The former phase-two positioning override is no longer needed.
- Width changes are included in martyria update detection so toolbar previews refresh when properties such as automatic mode, sign overrides, or scale affect the rendered result.
- The LaTeX schema version comment now explicitly documents that only breaking changes require a version bump.

## Reviewer Notes

- Phase-one right-aligned martyria geometry, especially paragraph-start elements and combinations containing both an attached fthora and a right tempo sign.
- Ink-bound calculations and whether the rendered spacing remains consistent across supported neume fonts.
- Analysis ordering: martyria fthora, quantitative interval, then quantitative-neume fthora.
- Save/load compatibility for scores created before the optional fields existed.
- LTR and Melkite RTL behavior around terminal ink reservations.

## Testing

- Full Vitest suite: 501 passed, 6 skipped.
- ESLint and Prettier checks passed.
- Added unit coverage for:
  - Save/load round trips
  - Analysis ordering and scale changes
  - Phase-one right-margin alignment
  - HTML export spacing and fthora output
- Manually verified in the browser that:
  - The toolbar targets the correct component.
  - Repeated fthora clicks cycle through top, bottom, and removal.
  - The martyria preview updates after property changes.
  - The right edge remains unchanged when an attached fthora is added.
- `vue-tsc --noEmit` was attempted but still encounters the pre-existing `reka-ui`/`@internationalized/date` `DayOfWeek` dependency mismatch.


basil (dismissed):
I don't have time to review this in detail but I have no objections here. I am mildly curious to see a real example of when this would be needed, because I can't think of one off the top of my head, but the use case seems reasonable enough.

danielgarthur:
The only thing that immediately comes to mind is the plagal second praises:
image

One could write the ending hard chromatic pa but then immediately follow it with the neume bringing you to Di, then put the fthora on that. Seems reasonable to me.

image

PeTH2:
> The only thing that immediately comes to mind is the plagal second praises: image
> 
> One could write the ending hard chromatic pa but then immediately follow it with the neume bringing you to Di, then put the fthora on that. Seems reasonable to me.
> image

Similar figures get used in some handwritten mixed-mode and makam compositions, like this setting of the Great Doxology in Defteroprotos, which I've been wanting to type up for a little while.
Screenshot_20260903_120157

basil:
> Similar figures get used in some handwritten mixed-mode and makam compositions

Yes, come to think of it, I have seen this in some initial martyriae for makam-based compositions, like this one for Makam Karcığar, with the note being sometimes placed in parentheses:

image

PR #2655, opened 2026-09-11
feat: centering for long melismas

### LINKED ISSUE #2547: Melisma edge case
The software currently produces this:

Image

I don't think this looks too bad, but I'm wondering if it would look better if the apostrophos and yporroe were centered above the word "through" rather than left-aligned. This is in the unusual case that a syllable is longer than an entire melisma run, which is rare but possible in English. I don't think there's much past precedent here in historical publications, this is fairly uncharted territory. So what do you think? Would it look better centered? Any downsides to doing so?

danielgarthur:
Fixes #2547. Does not alter hyphenated melismas. Maybe it should. But I wanted to have this reviewed first.

## Summary

Centers long melismatic lyrics beneath the quantitative neumes they span while keeping line breaking, neighboring lyrics, vareia, punctuation, and measure bars correctly aligned.

### Rules, concessions, and assumptions

- Centering applies only to left-to-right melismas made from adjacent notes.
- Underscore melismas are centered only when the remaining space is too short to draw the melisma line.
- Hyphenated melismas are centered whenever the lyric is wider than its starting neume; visible hyphens are then positioned after the centered lyric.
- Vareia and measure-bar reservations are excluded from the neume span used for centering.
- Punctuation positioning is preserved and included when determining whether the lyric fits.
- Knuth–Plass may not break within the shortest neume prefix needed to contain the lyric. Explicit line and page breaks still take precedence.
- Glue inside that protected prefix is fixed at its normal collision-aware preferred width.
- Breaks remain available later in a longer melisma once the lyric-containing prefix has been satisfied.
- Melkite RTL melismas remain left-aligned.
- Any melisma containing a martyria, tempo, or inline text box remains left-aligned. These cases are intentionally excluded from the new centering behavior.

## Major Changes

- Centers qualifying underscore and hyphenated lyrics beneath their complete quantitative-neume segment.
- Treats the shortest lyric-containing neume prefix as an atomic run for automatic line breaking.
- Uses the same glyph-collision and measure-bar-aware boundary widths for both prefix calculation and Knuth–Plass glue.
- Preserves normal spacing between the vareia and the preceding neume instead of shifting the entire melisma group during final positioning.
- Prevents centered lyrics from colliding with preceding or following syllables.
- Preserves a complete melisma line beneath a preceding melismatic neume when the following centered lyric would otherwise project into that space.
- Resolves punctuation, vareia, and measure-bar geometry in a consistent quantitative-neume coordinate system.
- Removes the previous iterative layout loop and its convergence/fallback handling.
- Documents the single-pass encoding and its eligibility rules in `notes/knuth_plass.md`.

## Implementation Notes

Phase 1 precomputes eligible note-only melismas and determines the shortest quantitative-neume prefix wide enough to contain each lyric. Each internal boundary uses the same preferred structural width used by normal layout, including neume-glyph and measure-bar clearance.

Breaks inside the selected prefix receive `MAX_COST`, and the associated glue is fixed so the geometry used to select the prefix matches the geometry supplied to Knuth–Plass.

If the complete melisma is narrower than the lyric, its centered bounds are known before line breaking and are used immediately for collision spacing. Otherwise, the protected prefix contains the lyric, allowing Phase 2 to center it beneath the final line segment without changing its external collision envelope.

Mixed-element melismas are detected using a shared continuation traversal. This traversal also allows hyphenated lyrics to recognize continuation notes beyond intervening martyria, tempo, or inline text-box elements, ensuring those melismas receive the intended left alignment.


danielgarthur:
@basil This seemed surprisingly complicated in that it requires multiple layout passes. I would appreciate a second opinion. Also, do you think this should be done for hyphenated melismas, too?

basil:
I suppose it should be done for hyphenated melismas. Perhaps such a case could be programmatically found to verify it, since I can't think of a real one offhand.

I tested this locally and it doesn't seem to be working for the slow Prosomion "Having laid up all their hope":

image

It does seem to work on the fast "By conceiving":

image

Although now "hymned" is starting a bit underneath the preceding apostrophos. Maybe that is OK... but I seem to recall you didn't like that and had fixed a similar bug before, when looking at at similar case in the first line of Doxologies ("good will towards")?

src/services/LayoutService.ts
```diff
@@ -378,15 +394,75 @@ export class LayoutService {
...
+          centeredMelismaLyrics,
+          result.centeredMelismaLyrics,
+        )
+      ) {
+        break;
+      }
+
+      if (pass + 1 >= maxCenteredMelismaLayoutPasses) {
+        // A discrete line-break change can theoretically make centered
+        // geometry oscillate. Never return pages positioned with geometry
```
basil: I thought that was the problem I solved in https://basilcrow.com/blog/relaxing-knuths-restrictions/ by coming up with a way to encode break-dependent widths in standard Knuth-Plass algebra.

danielgarthur:
Apparently the last round of changes broke some stuff that I had working.

basil:
Your previous "good will towards" fix makes this look good:

image


But this version looks bad again (with a hyphen instead of word extension):

image

Ideally I think the "good will towards" fix would apply both to the above and to the below:

image

src/services/LayoutService.ts
danielgarthur: I thought something was off... I would think we could encode each potential breakpoint in a melisma with negative width to account for the lyric centering that would result if the break occurs. I'm going to try that.

danielgarthur:
Made it use one pass using negative glue widths and such. Had to make several concessions. If you think any are crazy let me know. 

- Melismas containing a martyria, tempo, or inline text box remain left-aligned.
- A melisma cannot automatically break before the lyric has enough neumes to fit beneath it. For example, if the lyric spans the first two neumes of a four-neume melisma, the break cannot occur after the first neume.
- Spacing between those required neumes is fixed and will not stretch or shrink during justification.
- Manual line and page breaks still take precedence.
- Once the lyric has been fully accommodated, the remainder of a longer melisma may break normally.

danielgarthur:
Not ready for review yet. Moving to draft.

danielgarthur:
To summarize the reasoning behind the constraints:

Suppose a melisma has four neumes, and its lyric needs the first two to fit:

```
N1—N2 | N3 | N4
```

If we break after N2, then the lyrics should be centered under `N1—N2`. But if we don't, then the lyrics should be left aligned under `N1—N2 | N3`. Without treating `N1—N2` as unbreakable group, this becomes impossible to represent through penalty and negative glue. At least as far as I could figure out. That's not to say the KP algorithm couldn't be further modified, but it seems reasonable that we probably want to keep `N1—N2` together anyway.

I kept the penalty on `N3` to encourage `N3` to stay on the same line as `N1—N2`.

For melismas that pass through tempos or martyrias, it was unclear how those should work. I figured a tempo or martyria encountered should force a left alignment, but  it was too difficult to figure out whether a line would contain them ahead of time, and it's an edge case anyway. So I left those always left-aligned, which I think most of the time they would be left aligned anyway.

danielgarthur:
I still need to peruse through more scores, but here are some samples.

image


image


image


image

basil:
That reasoning makes sense to me. If I'm understanding correctly, that would result in this break being now banned:

image

I always felt such breaks were ugly and wanted to discourage them more, if not to ban them. Whatever cost we might eat in terms of a higher adjustment ratio, the above looks even worse, so this seems like a worthwhile tradeoff to me.

---

The updated examples look really great. I have one minor visual suggestion, which you can ignore if it makes things too complicated. In cases like this, we're already in a somewhat pathological situation trying to balance opposing concerns of tightness with readability. I think we can squeeze a little more tightness out of e.g.

image

by pushing "hymned" right up to the edge of the note boundary:

Untitled

My reasoning here is, we're already taking the hit of the extra space, let's find a way to minimize the hit we have to take without compromising readability. Readability is only really compromised when "hymned" directly clashes with the previous note, so we can edge a little closer. The underscore would have to get a little shorter, but that is a worthwhile tradeoff to me, and it already behaves that way in other spacing scenarios in the app, so this isn't introducing a new inconsistency.

This is a bit of a subjective opinion, but so is this whole feature, so I would be OK if you had a different opinion, or if this was too complicated to support.

danielgarthur:
Good catch. The spacing you described is how it works for long centered lyrics that are not melismas. So this was a bug when centering long melismas. It's been fixed.

basil:
The output is looking good to me. I took a clean up pass with Claude; feel free to revert if you disagree with it. Tested before and after cleanup pass and the examples look good to me

---

Before commit 4d5c29f3, the broad single-pass architecture was correct, but several mismatches remained between the geometry encoded during Phase 1 and the centering performed during Phase 2.

## 1. It handles non-monotonic spans caused by negative glue

Previously, atomic-prefix discovery stopped at the first note where the quantitative-neume span contained the lyric. That assumed the cumulative span could only increase.

Negative spacing can violate that assumption. For example:

```text
after N1: 40px  too short
after N2: 65px  fits
after N3: 55px  too short again
after N4: 80px  fits
```

The previous implementation could allow a break after `N3`, even though that segment was too narrow for the lyric. Commit 4d5c29f3 examines every possible segment end and protects every boundary through the last deficient note. This preserves the invariant that every permitted automatic segment end contains the lyric.

This is directly related to the generalized Knuth-Plass encoding: the line breaker supports negative widths, so prefix discovery must not assume cumulative widths are monotonic.

## 2. It centers lyrics before positioning dependent hyphens and underscores

Previously, a long melisma lyric was centered inside the underscore-layout loop. A preceding underscore or hyphen could already have been positioned against that lyric's old left-aligned position.

Commit 4d5c29f3 runs `centerLongMelismaLyrics` before laying out any hyphens or underscores. All melisma marks therefore use the final lyric positions. When a leading line-start hyphen belongs to a centered lyric, its offset is shifted by the same amount as the lyric.

This is an additional Phase 2 traversal, not another complete layout pass: it does not rerun Knuth-Plass or reposition the notes.

## 3. It separates the natural melisma end from the following-lyric clamp

Previously, the underscore was first shortened to clear the following lyric, and the shortened width was then used to decide whether the starting lyric should be centered. A naturally long underscore could therefore trigger centering merely because collision handling shortened it below `lyricsMelismaCutoffWidth`.

That created a circular dependency: the following lyric affected the centering decision, while centering could itself change the following-lyric geometry.

Commit 4d5c29f3 calculates the unclamped underscore end and the following-lyric limit separately. It decides whether to center from the unclamped end, then applies the clamp only when drawing the underscore. An underscore hidden solely because it was shortened to clear the next lyric now leaves its starting lyric left-aligned.

## 4. It permits Phase 2 centering only when Phase 1 made it safe

Centering moves a long left-aligned lyric toward the following syllable. If another lyric follows on the same line, that movement can require additional spacing which must be represented during line breaking.

Phase 1 already records group endpoints for which it reserved `lyricsMinimumSpacing` before the following lyric. Commit 4d5c29f3 passes that information into Phase 2 and centers a same-line segment only when its endpoint has the required reservation.

This prevents Phase 2 from introducing a lyric collision that the Knuth-Plass input did not account for.

## 5. It distinguishes centering eligibility from actual Phase 1 centering

Previously, the special zero-base inter-note width could be selected when the next melisma was merely eligible for centering. Eligibility did not guarantee that Phase 1 had actually centered the lyric beneath a deterministic group.

Commit 4d5c29f3 adds `phase1CenteredStartNotes`, which contains only melisma starts whose complete, fixed group was narrower than the lyric and whose centered bounds were therefore known before line breaking. The special zero-base spacing is now used only for those actual Phase 1 centers.

This keeps the spacing encoded for Knuth-Plass consistent with the geometry that Phase 2 will produce.

## 6. It avoids leftward Phase 2 movement that was not reserved

Phase 1 keeps preceding content clear of the lyric position it encodes. If Phase 2 subsequently moved an otherwise left-aligned lyric farther left, it could collide with the preceding lyric or a line-start hyphen.

Commit 4d5c29f3 performs new Phase 2 centering only when it moves the lyric right. A lyric whose centered position would move left remains at its Phase 1 position. Cases that Phase 1 itself centered remain safe because their centered outer bounds were already included in the line-breaking geometry.

## 7. It excludes Greek melismata from underscore-specific geometry

When Greek melismata are enabled, the renderer uses repeated vowels rather than an underscore. Applying atomic underscore-centering rules to those melismas could alter line breaking and lyric offsets for geometry that Phase 2 would never render.

Commit 4d5c29f3 excludes those Greek melismas from the atomic centering calculation. Greek melismata keep their existing repeated-vowel layout, while Greek lyrics still use underscore centering when Greek melismata are explicitly disabled.

## 8. It puts the post-prefix penalty beside the related melisma-start penalty

The first permitted break after a multi-neume atomic prefix remains discouraged so that the opening lyric-bearing group is not unnecessarily isolated. Commit 4d5c29f3 moves this cost into `getBreakCost`, alongside the equivalent penalty immediately after an ordinary melisma start.

This gives both cases the same explicit-break handling and keeps all graded melisma-edge break costs in one place.

## Overall effect

Commit 4d5c29f3 strengthens the contract that makes the single-pass design correct:

- Phase 1 protects every unsafe automatic endpoint and reserves the external geometry that centering may require.
- Phase 2 makes its centering decision from stable, unclamped geometry.
- Phase 2 centers before positioning dependent marks.
- Phase 2 declines changes that Phase 1 did not make safe.

The result is a more deterministic agreement between line breaking and final positioning.

basil:
> Does not alter hyphenated melismas. Maybe it should.

I suppose it should, or at least I can't think of any reason why it shouldn't.

danielgarthur:
I would need to see concrete examples of  hyphenated melismas that currently look bad. I think because of the hyphenation the left aligned approach ends up saving more space.

basil:
> I would need to see concrete examples of hyphenated melismas that currently look bad.

See the Automelon "The power of Thy Cross"

image

image

danielgarthur:
Updated.

image

basil:
Many thanks, this all looks great to me visually.

basil (dismissed):
I put this through Claude and Codex reviews and fixed any review feedback they gave, then manually tested again. All looks good to me. Thanks again for working on this, I think it looks really great! I also rebuilt all of the scores on my web site with this PR in case you want to do any further spot checks.

PR #2658, opened 2026-09-15
feat: implement spaceAfter as hskip

danielgarthur:
This PR re-imagines `spaceAfter` as a TeX hskip.

### Implementation Details

- spaceAfter remains unchanged in the UI, model, and save format.
- Notes, martyrias, and tempos now emit separate fixed glue with positive or negative width and zero stretch/shrink.
- The glue is discarded at line and paragraph breaks.
- Element widths no longer include spaceAfter.
- Removed the advance/ink-overhang distinction and related helpers.
- Preserved the martyria stale-width fix via getMartyriaBoxWidth.
- Updated the Knuth–Plass documentation.

### User-facing changes

The visible change is that `spaceAfter` now behaves like TeX `\hskip`, rather than enlarging or shrinking the preceding element.

- Between elements on the same line, `spaceAfter` still moves everything following the note, martyria, or tempo by exactly that amount.
  - Positive values add fixed space.
  - Negative values pull following content left.
  - It never stretches or shrinks during justification.

- At a line break immediately after the element, `spaceAfter` disappears. It no longer reserves blank space at the end of the preceding line.

- At the end of a paragraph, `spaceAfter` also disappears. In particular, it no longer offsets a terminal or right-aligned martyria from the right margin.

- Line-breaking decisions can change. A positive `spaceAfter` counts when two elements remain on the same line, but does not count when the breaker chooses the boundary after it. Conversely, a negative value no longer makes the preceding box artificially narrower at a line ending.

- Automatic lyric, glyph-overhang, and measure-bar collision spacing is calculated independently of `spaceAfter`. The manual adjustment is applied afterward:
  - Positive values add their full amount beyond the automatically safe spacing.
  - Negative values can deliberately pull elements back into that spacing and create overlap.

- The note, martyria, or tempo’s own geometry now ends at its visible/notated content. `spaceAfter` is no longer part of its width, so selection/layout diagnostic bounds and width-dependent placement no longer include invisible trailing space.

The `spaceAfter` control and saved document property are unchanged; only its layout semantics change.

Closes #2580


danielgarthur:
@basil I submit this alternative to removing `spaceAfter` for your consideration. I believe this would retain any advantage that the feature may have, while making the maintenance overhead almost nothing.

basil:
LGTM, but are only positive values supported or negative values as well? If only positive then should we clamp the UI to positive values?

danielgarthur:
Both positive and negative are supported.

basil (approved):
Makes sense, LGTM

PR #2666, opened 2026-09-16
fix: remove outdated font fallback from rich text

danielgarthur:
## Summary

Remove obsolete neume-font fallbacks from rich-text font stacks while preserving correct neume rendering through dedicated widget styles.

## Motivation

Rich-text neumes now have their own font-family variable and no longer need `Neanes` appended to every text font. See #2451. Keeping that fallback polluted serialized font-family values and coupled ordinary text styling to neume rendering. This caused spaces to disappear when a user opened a file that used a font that was not installed on the computer.

## Major Changes

- Store and serialize only the selected text font from CKEditor and the properties pane.
- Remove neume fallbacks from default fonts, paragraph styles, explicit inline styles, and HTML export text styles.
- Continue rendering editor neume widgets through `--ck-content-neume-font-family`.
- Add equivalent scoped neume-widget rules to standalone HTML exports.
- Preserve inherited text styling for plagal Greek letters.
- Remove obsolete fallback helpers and related code paths.

## Implementation Notes

- Standalone HTML uses `--byz-neume-font-family` for `.neanes-ck-neume`.
- Exported alignment and plagal layout rules mirror the editor’s widget CSS.
- Existing saved inline styles containing `,Neanes` are not migrated automatically; applying a font again serializes the simplified value.


basil (dismissed):
LGTM. Agents claimed that "right-aligned neumes jump to the page corner in exported HTML":

> `richTextParagraphStyleCss.ts:16` still sets `position:absolute!important`, and nothing positions `.byz---rich-text-box` or the elements around it (`ByzHtmlExporter.ts`:436-446), so those neumes end up in the document's top-right corner.

PR #2669, opened 2026-09-17
fix: use consistent baseline for lyrics

### LINKED ISSUE #2668: Baseline inconsistency
In my setting of the Akathist Canon, I follow a convention (originally seen in the Music Appendix to HTM's 1991 edition of this translation) where the words in italics at the end of the final troparia of the Third, Fourth, FIfth, Sixth, Seventh, and Eighth Odes indicate that these final words should be chanted by the opposite choir. I like this convention as it is typographically subtle yet still informative without the overhead of an explicit rubric. However, Neanes renders the italicized portion a bit lower than the normal portion:

Image

danielgarthur:
Fixes #2668.

## Summary

Aligns lyrics to a consistent baseline when neighboring notes use different font families, sizes, weights, or styles. It also corrects lyric and drop-cap alignment at non-default zoom levels and adds a developer overlay for inspecting lyric baselines.

## Motivation

Lyrics were previously positioned using a shared top coordinate. Because fonts have different ascent metrics, mixed lyric styles produced visibly different baselines.

Browser zoom introduced another discrepancy: Chromium can round displayed font metrics differently from simply scaling the canonical measurements, causing the app preview to differ slightly from PDF output.

## Major Changes

- Establishes a canonical lyric baseline from the page setup and default lyric font.
- Positions each lyric using its own font ascent while preserving the shared baseline.
- Measures and caches vertical font metrics at the displayed size to reduce zoom-related rounding errors.
- Applies the shared lyric baseline correction to drop caps without disturbing their naturally scaled CSS line boxes.
- Adds a “Show lyric baselines” developer-panel option that draws the baseline across each score line.
- Aligns the diagnostic stroke with the same painted edge used by melismas.

## Implementation Notes

- Canonical layout remains canvas-based and continues to drive PDF output.
- Displayed lyric geometry is recalculated from the actual zoomed font size rather than multiplying canonical ascent and height measurements.
- Drop caps receive only the common lyric-baseline translation. Independently remeasuring their large CSS line boxes introduced greater canvas-versus-CSS rounding error.
- Print mode uses a zoom factor of 1, preserving the canonical PDF layout.
- Font metrics are cached within the displayed metrics context to avoid repeated canvas measurements.

## Testing

- Manually inspected mixed lyric styles, melismas, the developer baseline overlay, and drop-cap alignment at multiple zoom levels in Electron, and exported to PDF.

## Known Limitations

Canvas and CSS inline layout use slightly different font-metric rounding paths. Small subpixel differences may remain for unusual fonts or extreme zoom levels. Eliminating them completely would require cached DOM baseline measurement.


basil:
Pushed commit 316f2d3e773ae04e63c4286cd282467c26a213e0 with agent feedback and manually tested, works well for me

danielgarthur:
Thanks!

basil:
Design problem mentioned by agents:

> 1. src/services/LayoutService.ts:3134 — adjustDropCapPosition still derives multi-line drop cap alignment from the constant neumeLineHeight, but getLineHeight can now return a taller resolvedNeumeLineHeight. A 2-line drop cap whose second line carries an oversized lyrics style under-estimates distanceFromTopToBottomOfLyrics, so the cap's bottom misses the lyrics baseline.
> 2. src/services/LayoutService.ts:1616 — same stale constant in the page-fit reservation (neumeLineHeight * dropCapContinuationLines). With grown continuation lines, requiresNewPage under-reserves and the paragraph overflows past innerPageHeight.
>
> Drop caps are positioned at 1729 in the same pass that resolves line heights, but a continuation line's height isn't known until that line is laid out. Fixing this properly means either a second pass over drop caps after heights settle, or propagating resolved heights forward. This is the one real design problem in the branch.

danielgarthur:
My agents flagged the same thing, but I do not believe it's accurate. Increasing the font size does not increase the line height. When I pushed back, the agents said "oh yeah you're right" (not that that means anything). But I am unaware of any mechanism by which this could be true.

danielgarthur:
I stand corrected. If I make the font stupid large, then yes, the line height changes... but by the time it does the font will have collided with the neumes or at least be so large that I can't imagine any legit use case for such a size.

I don't think this is a case worth considering unless it's easy to address.

basil (dismissed):

danielgarthur:
@basil I went ahead and attempted to address this. Not so much for the sake of drop caps but for the sake of page breaking. I confirmed that the below statement was true and that this commit fixes it.

> Expanded lyric lines are not repaginated correctly
LayoutService.getLineHeight() can now increase a completed neume line based on per-note lyric metrics. However, pagination is checked only after the next line has been appended.
If the completed line’s expanded height crosses the page boundary, the code pops the newly appended line and leaves the oversized previous line on the old page. That line can overlap the footer or run off the page. The document’s final line is never finalized, so it has the same problem without needing a following line.

basil (approved):

Issue #2624, opened 2026-09-02
Improve space before and after music paragraphs


basil:
I've been slowly building a book with NeanesTeX, and one of the big problems I've had is spacing before and after score snippets. Horizontal spacing is handled very well already by Knuth-Plass, but vertical spacing still leaves a lot to be desired. I started by trying to just fix NeanesTeX to be consistent with what Neanes itself produces, but then I realized that I don't like that either, so opening the discussion to see if we can make that more sensible first.

<img width="1480" height="1806" alt="Image" src="https://github.com/user-attachments/assets/d1952bd6-9801-45ed-8ce0-0fece5eda417" />

Here I just have a plain score paragraph with text boxes before and after it. The spacing seems fairly arbitrary, and not pleasing. For example, at the top, the music is way too close to the text. At the bottom, there is a more reasonable amount of space between the music and the text. Why are these different, and should they not be made consistent with each other. It would be very strange in Microsoft Word if you inserted a picture and there was more space below it than above it (with 0pt space before and after settings). I think these should likely be consistent.

The other concern I have is, what exactly should that space value be? Because here we have a boundary between a music paragraph, with its own leading (governed by page setup), and text with its own leading (configured in properties). So I think the space before the first line of music and the space after the last line of music in a musical paragraph should be defined by the music's leading, not the leading of the text before and after it. The screenshot I showed above seems to define the space somewhat arbitrarily, and possibly by the text's leading rather than the music's leading. Think of it this way: if you had 3 paragraphs in Microsoft Word, and the 1st and 3rd were in 12-pt font, but the second was in 36-pt font, you would want the second line to have more leading:

<img width="462" height="270" alt="Image" src="https://github.com/user-attachments/assets/0224abbc-d395-483d-b94e-a188a9fef87d" />

In the above example, the highlighted "S" has leading above and below it matching the size of the font. It would be strange indeed if it had the same leading as the text before and after it in 12-pt font. The bigger font requires more leading.

I'm happy to fix NeanesTeX to match Neanes itself, but I think Neanes itself should be made a little more consistent and sensible before I do that.

basil:
One of the places where you can clearly see this issue is if you include two score paragraphs back to back in NeanesTeX (in two separate include statements). The space between each score paragraph is lower than the space between each line of the score paragraphs. This isn't a problem in Neanes itself. As mentioned previously, NeanesTeX could be bugfixed to match Neanes, but I think a deeper design discussion is warranted about how this "should" work in both contexts.

[...]

PR #2674, opened 2026-09-20
Leading for music

basil:
### Problem

Music-line spacing and page breaks account for unused font space as if it were visible content. Extra space is not distributed evenly above and below the music, and a line can move to the next page even when its content would fit.

### Evaluation

The previous layout used a single height per line, expanded to include the full lyrics font box, and accumulated that entire height when checking page capacity. It did not distinguish content from leading (the extra space around content), or measure the actual lyric descenders on each line.

### Solution

Distribute spare music-line height evenly above and below the content, using the neume ascender and actual lyric descent as its vertical limits. Omit leading above the first line and below the last line on a page so pagination reflects the space the content needs. Add developer guides for neume ascenders and lyric descenders to make these boundaries visible.

### Implementation

- Replace scalar line heights with content height, element offset, and leading before and after each line. Text blocks, mode keys, and standalone images retain fixed content metrics.
- Measure and cache lyric advance width and ink descent together, respecting capitalization variants. Include drop-cap descent on the line whose lyrics baseline it shares.
- Use the new metrics for page placement, multiline drop-cap alignment, and keeping drop-cap continuation lines together when possible. Inline images extend music-line content while retaining the music ascender origin.
- Add disabled-by-default developer overlay preferences and account for font measurement differences at the displayed zoom, while retaining canonical measurements for print.


basil:
@danielgarthur My first attempt at a clean model for #2624, inspired by how CSS models leading. As-is, I think this diff is a meaningful improvement from the status quo.

The text box in the example below is an equal distance from the music above and below it:

image

Also every page starts at the ascender:

image

I've also cleaned up the notion of the "lyric descender" to mean "the lowest actual descender," so that really large drop caps that don't actually have e.g. a lowercase 'y' don't reserve space for it.

---

Things that still would need to be done:

(a) Make NeanesTeX consistent with this model and implementation. That is easy enough to do, once we agree on the final form of this PR.

(b) As of these changes I feel pretty comfortable with the definition of the "bottom" of the music line (lowest descender of all actually printed lyric or drop cap descenders). The latter can stretch if needed beyond the "line spacing" defined in page setup. All of that seems good and proper to me. But the corresponding definition of the "top" of the music line doesn't feel as well defined. I'm just using the concept of the music font ascender. It feels like this should be able to stretch upward somehow, just like a really big descender can stretch downward. For example, it feels like it should be able to stretch upward in some fashion if there is a really high note with a gorgon and isokratema, etc. But I'm not sure if that should be a part of this PR or would be too invasive of a change.

basil:
To add to the above, while I think "ascender" is the right concept overall, it seems to be defined in a strange way in this music font, where for a "typical" line, the ascender is defined above the tallest isokratema (?)

image

danielgarthur:
In EZ Psaltica, the ascender is just above the highest point of the highest ypsili that can appear. However, I don't think we should use the font ascender for this case. I think we should use the actual ink bounds and add some clearance between the top of the bounding box and whatever is above it. How much clearance, I couldn't say offhand, but I'm sure we can come up with something reasonable.

The analogous case in Western notation would be a staff that supports ledger lines. If a particular line of music doesn't contain any notes above the staff, you wouldn't reserve vertical space for the highest ledger line that could theoretically appear. You would leave some normal clearance above the staff. If ledger-line notes are present, the required space grows with the actual notation, with some clearance above the highest note.

I think the same principle should apply here: allocate vertical space based on the actual ink that is present, plus an appropriate clearance, rather than the maximum extent that the font could theoretically contain.

basil:
The ink bounds makes sense to me at a high level and I guess that's what I'm already using for the lowest lyric descender. However one concern I have with the ink bounds approach (and I guess that applies to using them both for lower and upper bounds) is that I generally want a stable distance from ascender to ascender or baseline to baseline, with extra space only being taken in exceptional cases, like very large lyrics, very high isokratema (the equivalent of "ledger lines" etc). So measuring the distance from the highest ink bound to the lowest ink bound, subtracting that from the line spacing in page setup, and dividing that into a top leading and a bottom leading all sounds good in theory, but in practice that would make leading uneven I would think.

danielgarthur:
I see your point. I think then we can either use the font's ascent, or define a custom ascent that we add to the metadata. As to what that custom ascent would be, I think it's either the top of the highest ypsili plus some extra clearance space, or the top of the highest ordinarily encountered ypsili plus some clearance space. Probably the octave leap is the highest (at least in EZ Psaltica). The higher leaps are too rare to consider, I think. Then we can use ink bounds to increase space if the bounds are higher than the ascent.

Some images for reference:

image

image

basil:
> I see your point. I think then we can either use the font's ascent, or define a custom ascent that we add to the metadata. 

Per https://glyphsapp.com/learn/vertical-metrics there are a variety of fields and conventions that can be used, including some that define a separate vertical maximum and clipping value, which may fit our use case here.

> As to what that custom ascent would be, I think it's either the top of the highest ypsili plus some extra clearance space, or the top of the highest ordinarily encountered ypsili plus some clearance space. 

I see no reason why clearance space should be encoded in the ascender. The same web page describes a line gap field that seems like a better fit for such information. In any case, this program stores leading amount as an absolute value, not a multiple of the line height, and I am not changing that in this PR. In other words, our software defines the clearance space as an application-level setting and doesn't rely on the font at all for this purpose.

basil:
@danielgarthur I have filed https://github.com/neanes/neanes/issues/2680 to track the issue about adding additional space between music lines for exceptionally high lines and https://github.com/neanes/neanes/issues/2681 to track the issue about the default ascender being too high. These are real problems, but I have no interest in fixing them at the present time. I am trying to unblock my book's house style, which motivates me to fix only https://github.com/neanes/neanes/issues/2624 in this PR.

---

I checked how InDesign and TeX handle a line that is close to the bottom of the page, where its baseline would fit but its descender would not. Both of them keep the line on the bottom of the page, letting the descenders spill into the margin. This is the case regardless of whether or not the line actually contains e.g. a 'y' or 'g' character. My first attempted reserved the full lyric font box, including its descent, when deciding whether a music line fit on the page. Since I am ultimately aiming for compatibility with TeX and page layout applications, I chose to follow the same convention. As such I have partially reverted  the lyric font box height expansion added in #2669 (sorry!) as well as the ink bounds logic from my first attempt. commit fe7b02b6 judges the bottom of an ordinary music line only by the lyric baseline, without reserving space for lyric descenders.

I also found that I had inadvertently introduced a minor regression in #2215, which introduced a global Tailwind `line-height: 1.5`. This replaced the neume box's font based normal line height with a fixed multiplier. I have fixed that regression in commit fe7b02b6, alongside the layout changes described below.

## Compared with `origin/master`

- **Restores font-based neume rendering.** `.neume-box` now explicitly uses `line-height: normal`, overriding the inherited Tailwind 1.5 line height. The Page Setup neume preview also uses `line-height: normal`.
- **Changes music-line spacing.** Previously, each line occupied one height, enlarged as necessary to accommodate lyrics. Now the content runs from the neume font's ascender to the default lyrics baseline. Any remaining configured line height is split equally above and below that content ("half-leading").
- **Changes pagination.** A music line loses its top half-leading only when it is the first line on the page; a music line following a text box or other content retains it. The last line's bottom half-leading is excluded from the page-fit calculation. Consequently, more lines may fit on a page. This is consistent with TeX and InDesign.
- **Stops individual lyrics from expanding lines.** Custom lyric sizes and descenders no longer increase the reserved height. Pagination uses the default lyrics baseline, so unusually large lyrics can extend beyond the reserved content. This is consistent with TeX and InDesign.
- **Updates drop-cap positioning and page-fit calculations** to account for the new spacing.
- **Updates developer diagnostics.** Adds half-leading and neume-ascender guides, and aligns ink/collision overlays using the measured font ascent.
- **Text boxes, mode keys, and images** retain their existing height calculations, although their spacing relative to adjacent music lines changes.

## Compared with before #2215

#2215 introduced a global Tailwind `line-height: 1.5`. The existing `.neume-box` had no override. commit fe7b02b6's `normal` override restores the earlier font-based CSS behavior. However, commit fe7b02b6 does not simply restore the old layout. Before the port, music-line height was effectively:

```text
max(configured line height, neume font height + lyric offset + lyric font height)
```

Elements started at the line's origin, with spare space effectively following the content. There was no symmetric half-leading or trimming at page boundaries.

commit fe7b02b6 instead uses:

```text
content height = neume font height + lyric offset + default lyric ascent
half-leading = max(0, configured line height - content height) / 2
```

For example, with a configured height of 100 and content height of 70, commit fe7b02b6 assigns 15 above and below. Consecutive ordinary music lines still start 100 apart, but when a music line starts the page, its neume ascender starts directly at the usable page top. For ordinary music lines, the final default lyrics baseline determines page fit; multiline drop caps also reserve continuation space. A line placed alone on a page is retained even if it overflows. In other words, commit fe7b02b6 combines a restoration of pre-Reka CSS behavior with a new vertical-spacing and pagination model.

---

I'm looking for confirmation that this direction and these changes are desired / approved. If so, I'll then continue working on this changeset to make NeanesTeX apply and respect the same metrics / philosophy.

danielgarthur:
Yes, this approach makes sense to me.

One thing I noticed is that it leads to a larger gap between initial martyria and the first line. So I think the spacing for that needs to be figured out, too, rather than letting the browser and the font's metrics determine the spacing. This has always been an issue, but is more obvious now with extra leading spacing.

Take this for example:

image

Lots of space is reserved on the bottom. More so than on the top. Presumably to accommodate something like this:

image

or this:

image

I'm not sure what the right answer is here, but I think it's worth fixing this, too. I would be happy to do the actual work myself, but I could use your help figuring out what the correct rule should be.

danielgarthur:
I think the biggest issue is simply that the distance from the Η to the top of the bounding box should be the same as the distance to the bottom. As to how much space should be reserved for neumes, I don't know.

image

basil:
I think there are two approaches here. I can either try to find the minimal adjustment on document load to maintain backward compatibility with vertical metrics of existing documents that have [rich] text boxes or initial martyria objects, or we can change the layout a bit more intentionally. For initial martyria objects I've always found the hard-coded space to be too much and wished that it was a small amount that could be increased manually rather than a hard-coded large amount, something like this:

image

